PR #32, opened 2023-10-27
fix: update deps and env var

### LINKED ISSUE #25: Fix typo in staging secret name
Maybe let's uniformize secrets's content

https://github.com/graasp/graasp-account/blob/0749ae42125524146d3b1c8352447fe05fd47268/.github/workflows/deploy-stage.yml#L52C43-L52C83

spaenleh:
- fix: update deps
- fix: update readme
- fix #25

### DIFF
--- a/src/components/common/UserSwitchWrapper.tsx
+++ b/src/components/common/UserSwitchWrapper.tsx
@@ -27,6 +27,7 @@ const UserSwitchWrapper = ({ ButtonContent }: Props): JSX.Element => {
       renderAvatar={renderAvatar}
       profilePath={HOME_PATH}
       redirectPath={GRAASP_AUTH_HOST}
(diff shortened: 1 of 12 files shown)


src/components/common/UserSwitchWrapper.tsx
```diff
@@ -27,6 +27,7 @@ const UserSwitchWrapper = ({ ButtonContent }: Props): JSX.Element => {
       renderAvatar={renderAvatar}
       profilePath={HOME_PATH}
       redirectPath={GRAASP_AUTH_HOST}
+      userMenuItems={[]}
```
spaenleh, line 30: Ideally this props would be optional so we do not have to set it if we do not need it.

LinaYahya (approved):

PR #77, opened 2024-01-22
fix: update dependencies and fix issues

spaenleh:
In this PR:
- update dependencies
- fixes a bug where the spanish language would not be displayed in the LanguageSelect if it was selected
- fixes a warning about missing keys in select options
- fixes an error that blob does not have a name property (AvatarSettings)


pyphilia (approved):
LGTM 🖼️

ReidyT (approved):

PR #89, opened 2024-02-08
fix: add timestamp in html

spaenleh:


ReidyT (approved):
LGTM 👍. I'm curious, what is the purpose of adding the build timestamp ?

spaenleh:
@ReidyT This allows me to know when the file was built for example to trouble shoot some cache issues etc, in general I think it is an indication that can prove useful in the future to investigate.

PR #165, opened 2024-03-19
fix: accept query params for validation social links

### LINKED ISSUE #88: Copy pasting a profile link from twitter shows up as wrong because it contains a query param
When copy pasting a profile link it may contain query params:
Screenshot 2024-02-05 at 17 00 02

if we remove the query params it is accepted as correct:
Screenshot 2024-02-05 at 17 00 08

Either:
- remove query params from url when pasting
- allow query params in valid urls

But this is not friendly to users that may not understand why their link is not accepted.

LinaYahya:
closes #88


spaenleh (approved):
Thank you for the fix. Is it true that this will allow query parameters to be used in the URL, but that they will be scrubbed out when saving (since we only save the social id) ?

PR #169, opened 2024-03-22
fix: set width of public profile form

### LINKED ISSUE #69: [UI Bug] Weird jump in content width on public profile page
There is a weird content jump bug in the public profile page in account:

https://github.com/graasp/graasp-account/assets/39373170/6bb1166d-7683-420d-b42a-bd36b251d1f2

### Expected behaviour
I would expect the fields to have a maximum width (to not be 50km long on large screens), but to not exceed the viewport width on small screens

LinaYahya:
closes #69

### DIFF
--- a/src/pages/PublicProfileScreen.tsx
+++ b/src/pages/PublicProfileScreen.tsx
@@ -143,7 +143,7 @@ const PublicProfileScreen = (): JSX.Element => {
   return (
     <Main>
       <Grid container spacing={3}>
-        <Grid item sm={12} md={6} lg={6}>
+        <Grid item sm={12} md={10} lg={8} xl={6}>
           <Box sx={{ mt: 1, mb: 3 }}>
             <Typography variant="h4" component="h1">
               {t('PUBLIC_PROFILE_TITLE')}


spaenleh (changes requested):
I can see how this improves the jittering a bit by adding more steps. But I was thinking of something simpler. 
Full width when screen is small, and then simply add a maxwidth. See video below:

https://github.com/graasp/graasp-account/assets/39373170/97dd463d-2721-46ab-9e60-67174f468164

src/pages/PublicProfileScreen.tsx
```diff
@@ -142,8 +142,8 @@ const PublicProfileScreen = (): JSX.Element => {
 
   return (
     <Main>
-      <Grid container spacing={3}>
-        <Grid item sm={12} md={6} lg={6}>
+      <Grid container spacing={3} sx={{ maxWidth: '700px' }}>
+        <Grid item xs={12}>
```
spaenleh, line 146: Is the `Grid` really necessary in this case ? Maybe it is possible to replace it by a simple `Box` ?

spaenleh (changes requested):
Nice, maybe we can even remove the Grid for a Box ?

spaenleh (approved):

PR #180, opened 2024-03-28
feat: allow to change username

mariembencheikh:
#29 

https://github.com/graasp/graasp-account/assets/58473159/f0705585-52c4-449d-b411-af3be804e07c

image

### DIFF
--- a/cypress/e2e/profile.cy.ts
+++ b/cypress/e2e/profile.cy.ts
@@ -1,6 +1,68 @@
-describe('template spec', () => {
-  it('passes', () => {
-    cy.setUpApi();
+import {
+  USERNAME_CANCEL_BUTTON_ID,
+  USERNAME_DISPLAY_ID,
+  USERNAME_EDIT_BUTTON_ID,
+  USERNAME_SAVE_BUTTON_ID,
+} from '@/config/selectors';
+
+import { BOB, MEMBERS } from '../fixtures/members';
+
+const changeUsername = (newUserName: string) => {
+  cy.get(`#${USERNAME_EDIT_BUTTON_ID}`).click();
+  cy.get('input[name=username]').clear();
+  // Find the input field and type the new username
+  cy.get('input[name=username]').type(newUserName);
+};
+
+describe('Change username', () => {
+  beforeEach(() => {
+    cy.setUpApi({ currentMember: BOB });
     cy.visit('/');
+  });
+
+  it('Username field connot be empty', () => {
+    changeUsername('validUsername');
+    cy.get('input[name=username]').clear();
+    cy.get(`#${USERNAME_SAVE_BUTTON_ID}`).should('be.disabled');
+  });
+
+  it('Username too long', () => {
+    const longUsername = MEMBERS.WRONG_NAME_TOO_LONG.name;
+    changeUsername(longUsername);
+
+    cy.get(`#${USERNAME_SAVE_BUTTON_ID}`).should('be.disabled');
+  });
+
+  it('Username too short', () => {
+    const shortUsername = MEMBERS.WRONG_NAME_TOO_SHORT.name;
+    changeUsername(shortUsername);
+    cy.get(`#${USERNAME_SAVE_BUTTON_ID}`).should('be.disabled');
+  });
+
+  it('Valid username can be saved', () => {
+    const validUsername = 'validUsername';
+    changeUsername(validUsername);
+    cy.get(`#${USERNAME_SAVE_BUTTON_ID}`).should('not.be.disabled');
+
+    cy.get(`#${USERNAME_SAVE_BUTTON_ID}`).click();
+
+    cy.wait('@editMember').then(({ request: { body } }) => {
+      expect(body.name).to.equal(validUsername);
+    });
+  });
+
+  it('Should not update the user name if canceling edit', () => {
+    changeUsername('validUsername');
+    cy.get(`#${USERNAME_CANCEL_BUTTON_ID}`).click();
+    cy.get(`#${USERNAME_DISPLAY_ID}`).contains(BOB.name);
+  });
+
+  it('Saves username after trimming trailing space', () => {
+    const usernameWithTrailingSpace = 'test  '; // Nom d'utilisateur avec espace à la fin
+    changeUsername(usernameWithTrailingSpace);
+    cy.get(`#${USERNAME_SAVE_BUTTON_ID}`).click();
+    cy.wait('@editMember').then(({ request }) => {
+      expect(request.body.name).to.equal(usernameWithTrailingSpace.trim());
+    });
   });
 });
--- a/package.json
+++ b/package.json
@@ -3,7 +3,8 @@
   "version": "2.4.0",
   "private": true,
   "contributors": [
-    "Kim Lan Phan Hoang"
+    "Kim Lan Phan Hoang",
+    "Mariem Bencheikh"
   ],
   "engines": {
     "node": ">=20"
@@ -13,8 +14,8 @@
     "@emotion/cache": "11.11.0",
     "@emotion/react": "11.11.4",
     "@emotion/styled": "11.11.0",
-    "@graasp/query-client": "3.0.0",
-    "@graasp/sdk": "4.2.1",
+    "@graasp/query-client": "3.2.1",
+    "@graasp/sdk": "4.5.0",
     "@graasp/translations": "1.25.3",
     "@graasp/ui": "4.15.0",
     "@mui/icons-material": "5.15.14",
--- a/src/components/main/EditingNameField.tsx
+++ b/src/components/main/EditingNameField.tsx
@@ -1,0 +1,99 @@
+import { ChangeEvent, useState } from 'react';
+
+import CloseIcon from '@mui/icons-material/Close';
+import DoneIcon from '@mui/icons-material/Done';
+import { IconButton, Stack, TextField } from '@mui/material';
+
+import { useAccountTranslation } from '@/config/i18n';
+import {
+  USERNAME_CANCEL_BUTTON_ID,
+  USERNAME_INPUT_FIELD_ID,
+  USERNAME_SAVE_BUTTON_ID,
+} from '@/config/selectors';
+
+const MIN_USERNAME_LENGTH = 3;
+const MAX_USERNAME_LENGTH = 30;
+
+type EditingUserNameFieldProps = {
+  name: string;
+  onSave: (newValue: string) => void;
+  onCancel: () => void;
+};
+
+const verifyUsername = (username: string) => {
+  const trimmedUsername = username.trim();
+  if (trimmedUsername === '') {
+    return 'USERNAME_EMPTY_ERROR';
+  }
+
+  if (
+    trimmedUsername.length < MIN_USERNAME_LENGTH ||
+    trimmedUsername.length > MAX_USERNAME_LENGTH
+  ) {
+    return 'USERNAME_LENGTH_ERROR';
+  }
+  return null;
+};
+
+const EditingUserNameField = ({
+  name,
+  onSave,
+  onCancel,
+}: EditingUserNameFieldProps): JSX.Element => {
+  const [newUserName, setNewUserName] = useState(name);
+  const { t } = useAccountTranslation();
+  const [error, setError] = useState<string | null>();
+
+  const handleChange = ({ target }: ChangeEvent<HTMLInputElement>) => {
+    const { value } = target;
+    setNewUserName(value);
+    const errorMessage = verifyUsername(value);
+    setError(
+      errorMessage
+        ? t(errorMessage, {
+            min: MIN_USERNAME_LENGTH,
+            max: MAX_USERNAME_LENGTH,
+          })
+        : null,
+    );
+  };
+  const handleSave = () => {
+    const errorMessage = verifyUsername(newUserName);
+
+    if (!errorMessage) {
+      onSave(newUserName.trim());
+    }
+  };
+
+  return (
+    <Stack direction="row" alignItems="center" spacing={1}>
+      <TextField
+        id={USERNAME_INPUT_FIELD_ID}
+        variant="standard"
+        type="text"
+        name="username"
+        value={newUserName}
+        error={Boolean(error)}
+        helperText={error}
+        onChange={handleChange}
+        autoFocus
+      />
+      <IconButton
+        type="reset"
+        onClick={onCancel}
+        id={USERNAME_CANCEL_BUTTON_ID}
+      >
+        <CloseIcon />
+      </IconButton>
+      <IconButton
+        id={USERNAME_SAVE_BUTTON_ID}
+        onClick={handleSave}
+        disabled={Boolean(error)}
+      >
+        <DoneIcon />
+      </IconButton>
+    </Stack>
+  );
+};
+
+export default EditingUserNameField;
--- a/src/components/main/UsernameForm.tsx
+++ b/src/components/main/UsernameForm.tsx
@@ -1,0 +1,58 @@
+import { useState } from 'react';
+
+import { ModeEdit } from '@mui/icons-material';
+import { Box, IconButton, Stack } from '@mui/material';
+
+import {
+  USERNAME_DISPLAY_ID,
+  USERNAME_EDIT_BUTTON_ID,
+} from '@/config/selectors';
+
+import { mutations } from '../../config/queryClient';
+import EditingUserNameField from './EditingNameField';
+
+type UsernameProps = {
+  member: {
+    id: string;
+    name: string;
+  };
+};
+const UsernameForm = ({ member }: UsernameProps): JSX.Element => {
+  const { mutate: editMember } = mutations.useEditMember();
+  const [isEditing, setIsEditing] = useState(false);
+
+  const handleSave = (newUserName: string) => {
+    editMember({
+      id: member.id,
+      name: newUserName,
+    });
+
+    setIsEditing(false);
+  };
+
+  const cancelEditing = () => {
+    setIsEditing(false);
+  };
+
+  if (isEditing) {
+    return (
+      <EditingUserNameField
+        name={member.name}
+        onSave={handleSave}
+        onCancel={cancelEditing}
+      />
+    );
+  }
+  return (
+    <Stack direction="row" alignItems="center" spacing={1}>
+      <Box id={USERNAME_DISPLAY_ID}>{member.name}</Box>
+      <IconButton
+        onClick={() => setIsEditing(true)}
+        id={USERNAME_EDIT_BUTTON_ID}
+      >
+        <ModeEdit />
+      </IconButton>
+    </Stack>
+  );
+};
+export default UsernameForm;
--- a/src/config/selectors.ts
+++ b/src/config/selectors.ts
@@ -18,3 +18,9 @@ export const DELETE_MEMBER_DIALOG_DESCRIPTION_ID = 'alert-dialog-description';
 export const APP_NAVIGATION_PLATFORM_SWITCH_ID = 'appNavigationPlatformSwitch';
 
 export const STORAGE_PROGRESS_BAR_ID = 'storageProgressBar';
+
+export const USERNAME_EDIT_BUTTON_ID = 'username-edit-button';
+export const USERNAME_SAVE_BUTTON_ID = 'username-save-button';
+export const USERNAME_CANCEL_BUTTON_ID = 'username-cancel-button';
+export const USERNAME_INPUT_FIELD_ID = 'username-input-field';
+export const USERNAME_DISPLAY_ID = 'username-display';
--- a/src/langs/ar.json
+++ b/src/langs/ar.json
@@ -16,6 +16,7 @@
   "MAIN_MENU_STORAGE": "تخزين",
   "SAVE_ACTIONS_TOGGLE_TOOLTIP": "قريباً!",
   "PROFILE_TITLE": "الملف الشخصي",
+  "PROFILE_MEMBER_NAME": "اسم المستخدم",
   "PROFILE_MEMBER_ID_TITLE": "هوية المستخدم",
   "PROFILE_LANGUAGE_TITLE": "لغة",
   "PROFILE_CREATED_AT_TITLE": "عضوّ منذ",
--- a/src/langs/de.json
+++ b/src/langs/de.json
@@ -16,6 +16,7 @@
   "MAIN_MENU_STORAGE": "Speicher",
   "SAVE_ACTIONS_TOGGLE_TOOLTIP": "Demnächst verfügbar!",
   "PROFILE_TITLE": "Profil",
+  "PROFILE_MEMBER_NAME": "Nutzername",
   "PROFILE_MEMBER_ID_TITLE": "Benützer ID",
   "PROFILE_LANGUAGE_TITLE": "Sprache",
   "PROFILE_CREATED_AT_TITLE": "Mitglied seit",
--- a/src/langs/en.json
+++ b/src/langs/en.json
@@ -16,6 +16,7 @@
   "MAIN_MENU_STORAGE": "Storage",
   "SAVE_ACTIONS_TOGGLE_TOOLTIP": "Coming soon!",
   "PROFILE_TITLE": "Profile",
+  "PROFILE_MEMBER_NAME": "Username",
   "PROFILE_MEMBER_ID_TITLE": "Member ID",
   "PROFILE_LANGUAGE_TITLE": "Language",
   "PROFILE_CREATED_AT_TITLE": "Member Since",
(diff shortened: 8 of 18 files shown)


src/langs/de.json
```diff
@@ -16,6 +16,8 @@
   "MAIN_MENU_STORAGE": "Speicher",
   "SAVE_ACTIONS_TOGGLE_TOOLTIP": "Demnächst verfügbar!",
   "PROFILE_TITLE": "Profil",
+  "PROFILE_MEMBER_NAME": "Nutzername",
+  "PROFILE_EDIT_USERNAME_BUTTON": "Speichern",
```
pyphilia, line 20: Thanks for adding all the translations. But actually we have a tool to handle that, so feel free to fill for languages you know, and let empty for languages you don't know^^

src/langs/es.json
```diff
@@ -16,6 +16,8 @@
   "MAIN_MENU_STORAGE": "Almacenamiento",
   "SAVE_ACTIONS_TOGGLE_TOOLTIP": "¡Muy pronto!",
   "PROFILE_TITLE": "Perfil",
+  "PROFILE_MEMBER_NAME": "nombre de usuario",
```
pyphilia, line 19:
```suggestion
  "PROFILE_MEMBER_NAME": "Nombre de usuario",
```

src/langs/it.json
```diff
@@ -16,6 +16,8 @@
   "MAIN_MENU_STORAGE": "Archiviazione",
   "SAVE_ACTIONS_TOGGLE_TOOLTIP": "In arrivo prossimamente!",
   "PROFILE_TITLE": "Profilo",
+  "PROFILE_MEMBER_NAME": "nome utente",
```
pyphilia, line 19:
```suggestion
  "PROFILE_MEMBER_NAME": "Nome utente",
```

src/langs/it.json
```diff
@@ -16,6 +16,8 @@
...
+  "PROFILE_EDIT_USERNAME_BUTTON": "registrare",
```
pyphilia, line 20:
```suggestion
  "PROFILE_EDIT_USERNAME_BUTTON": "Registrare",
```

src/pages/MemberScreen.tsx
```diff
@@ -56,14 +61,53 @@ const MemberProfileScreen = (): JSX.Element | null => {
...
+                  {newUsername !== member.name && (
+                    <Button
+                      variant="contained"
+                      color="primary"
+                      sx={{ my: 1 }}
+                      onClick={() => updateUsername()}
+                    >
+                      {t('PROFILE_EDIT_USERNAME_BUTTON')}
+                    </Button>
+                  )}
```
pyphilia, line 104: Since this feature is quite well encapsulated, I think you can create a new file with this component only.

src/pages/MemberScreen.tsx
```diff
@@ -56,14 +61,53 @@ const MemberProfileScreen = (): JSX.Element | null => {
...
             <Typography variant="h4" component="h1">
-              {member.name}
+              {t('PROFILE_TITLE')}
             </Typography>
+            <Grid container alignItems="center">
+              <Grid item xs={4}>
+                <Typography>{t('PROFILE_MEMBER_NAME')}</Typography>
+              </Grid>
+              <Grid item xs={8}>
+                <Typography>
```
pyphilia, line 87: You can remove the `Typography` tag.

pyphilia (changes requested):
I added some intermediary comments, I'm looking forward the changes for the layout and the validation!

src/langs/ar.json
```diff
@@ -63,4 +65,4 @@
   "PROFILE_DELETE_TYPE_CONFIRMATION_TEXT": "لتأكيد الحذف، اكتب \"{{text}}\" أدناه:",
   "DELETE_CONFIRMATION_VALUE": "يمسح",
   "DESTRUCTIVE_SETTINGS_DETAILS": "يرجى العلم أن الإجراءات الموضحة هنا قد يكون لها عواقب لا رجعة فيها. تأكد من مراجعة أي قيود بدقة قبل المتابعة، ولا تتخذ الإجراء إلا بعد التأكد."
-}
+}
```
spaenleh, line 68: You are missing a newline here. Configure your text editor as follows: 
```json
"files.insertFinalNewline": true
```

src/pages/MemberScreen.tsx
```diff
@@ -56,14 +61,53 @@ const MemberProfileScreen = (): JSX.Element | null => {
...
+                  <TextField
+                    variant="standard"
+                    type="text"
+                    value={newUsername}
+                    onChange={onInputChange}
+                    sx={{ mr: 2 }}
```
spaenleh, line 93: If this is used for spacing, we prefer to use a stack with the `spacing` property. Is is more flexible and lets you re-arrange element with ease, without needing to move the margin property to the first element.
```tsx
<Stack direction="row" spacing={2}>
  <Textfield
  ....
  />
  {
    ....
  }
</Stack>
```

spaenleh (changes requested):
I agree with the comments of @pyphilia especially the ones for the translations 😉 Looking forward to your next changes. 

Do not forget to:
- include a screenshot in the PR description. Even a very short video with the interaction would be a nice touch.
- add a reference to the issue you are fixing in the PR description so it can be auto-closed. Otherwise we have to do some manual cleanup 😕

src/components/main/EditUsername.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { mutations } from '../../config/queryClient';
+
+type UsernameProps = {
+  memberId: string;
+  memberName: string;
+};
+const EditUsername = ({ memberId, memberName }: UsernameProps): JSX.Element => {
+  const { t } = useAccountTranslation();
+  const { mutate: editMember } = mutations.useEditMember();
+  const [newUsername, setNewUsername] = useState(memberName || '');
```
spaenleh, line 26: This useState creates a state that does not react to external changes of the username. In the current situation it is not really an issue because the full component tree is re-created when the current member query is re-fetched. 
But be careful with such implementations. Once side-issue however is that if you are editing the name, and the current member query is refetched, you will loose your changes. 

There currently is no websocket feedback on member update, so if you have two windows open and change your name in one it will not propagate to the other windows unless you manually refresh them. Would be nice to have some feedback there maybe.

src/components/main/EditUsername.tsx
```diff
@@ -0,0 +1,98 @@
...
+  };
+
+  const cancelEditing = () => {
+    setNewUsername(memberName);
+    setNewUsernameError(null);
+    setIsEditing(false);
+  };
+  return (
+    <Stack direction="row" spacing={2}>
+      {isEditing ? (
```
spaenleh, line 55: You have this conditional multiple times in your JSX. Would it make sens to create a new component for the Editing name ? and then you can add a `if` and render the right component based on `isEditing` I think explicit ifs are clearer than using ternary expressions in JSX for some cases.

spaenleh (changes requested):
Nice changes, I have added some comments, please review them and let me know if you have questions. We are on the right path ! 👍

src/components/main/EditUsername.tsx
mariembencheikh: More details please, how it must be

src/components/main/EditingNameField.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const [error, setError] = useState<string | null>();
+  const verifyUsername = (username: string) => {
+    if (username.trim() === '') {
+      return 'USERNAME_EMPTY_ERROR';
+    }
+    if (username.length < 3 || username.length > 30) {
+      return 'USERNAME_LENGTH_ERROR';
+    }
+    return null;
+  };
```
spaenleh, line 30: Put this function outside of your component.

src/pages/MemberScreen.tsx
```diff
@@ -71,7 +71,7 @@ const MemberProfileScreen = (): JSX.Element | null => {
                 {t('PROFILE_MEMBER_NAME')}
               </Grid>
               <Grid item xs={8}>
-                <EditUsername memberId={member.id} memberName={member.name} />
+                <EditUsername member={member} />
```
spaenleh, line 74: Would it make sens to rename this component to `UsernameForm` ?

src/components/main/EditUsername.tsx
```diff
@@ -1,95 +1,48 @@
...
-            type="reset"
-            onClick={() => cancelEditing()}
-          >
-            <CloseIcon />
-          </Button>
-        </>
+        </Stack>
       )}
     </Stack>
   );
```
spaenleh, line 48: Refactor this to be:
```tsx
if (isEditing) {
  return <EditingComponent/>
}
return <DisplayComponent/>
```

I think it is clearer than using the ternary operator.

src/components/main/EditUsername.tsx
```diff
@@ -1,95 +1,48 @@
...
   };
 
   const cancelEditing = () => {
-    setNewUsername(memberName);
-    setNewUsernameError(null);
     setIsEditing(false);
   };
   return (
-    <Stack direction="row" spacing={2}>
+    <Stack direction="row" spacing={2} alignItems="center">
```
spaenleh, line 32: This `Stack` is not necessary I think.

src/components/main/EditingNameField.tsx
```diff
@@ -0,0 +1,68 @@
...
+        error={Boolean(error)}
+        helperText={error}
+        onChange={handleChange}
+      />
+      <Button onClick={handleSave}>
+        <DoneIcon />
+      </Button>
+      <Button type="reset" onClick={onCancel}>
+        <CloseIcon />
+      </Button>
```
spaenleh, line 63: If you do not want to have the space between the two buttons here, you can group them.

spaenleh (changes requested):

spaenleh:
@mariembencheikh Can you test that you can update a name even if there are some spaces at the end ?

src/components/main/EditUsername.tsx
spaenleh: @mariembencheikh 
Whenever you create a state that gets an initial value from the props you will face a potential de-synchronisation. This is because the state will not be updated if the prop changes. To synchronise the state with prop changes you usually have to add a `useEffect`. 

In our case, the props is actually coming from an API call, so this is data from the server. 
Imagine that the user opens 2 windows. He changes the username in the first window. In the second window, imagine there is a mechanism to notify of the data change on the server (this can be a websockets update). It would result in a new call to your component with the name prop having a new value. But the state would not get updates unless you add a `useEffect` that react to updates of the name prop and sets the state accordingly. 

This is for the theory.
In practice with our example, when the data from the server changes, or is refetched manually by using the react-query dev tools (little flower at the bottom of the screen), nearly all the component tree is re-created, and thus the state is re-created with the new initial value. I just wanted to raise your attention to the de-synchronisation issue, as this is a common mistake. 

The last part of my comment was suggesting potential improvements.
If we take the example with multiple windows (even multiple tabs), I think it would be nice for the updated username to be propagated all the windows and tabs. Can you think of a way to do this ? Would you like to work on this task ? I think it would add a great value for the user, seeing up-to-date data on all open windows without having to refresh the pages manually.

src/components/main/EditingNameField.tsx
mariembencheikh: I want to have the space between

src/langs/de.json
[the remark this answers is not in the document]
spaenleh: Also it says "save" and the constant says "edit" maybe it needs to be adapted.

src/langs/ar.json
spaenleh, line 68: Add a newline to this file please.

src/langs/en.json
```diff
@@ -62,5 +64,7 @@
   "GO_TO_MANAGE_ACCOUNT_PAGE": "Continue",
   "PROFILE_DELETE_TYPE_CONFIRMATION_TEXT": "To confirm the deletion, type \"{{text}}\" below:",
   "DELETE_CONFIRMATION_VALUE": "delete",
-  "DESTRUCTIVE_SETTINGS_DETAILS": "Please be aware that the actions outlined here have potentially irreversible consequences. Ensure you thoroughly review any restrictions before proceeding, and only take action once you are certain."
+  "DESTRUCTIVE_SETTINGS_DETAILS": "Please be aware that the actions outlined here have potentially irreversible consequences. Ensure you thoroughly review any restrictions before proceeding, and only take action once you are certain.",
+  "USERNAME_EMPTY_ERROR": "The field cannot be empty",
+  "USERNAME_LENGTH_ERROR": "Username must be between 3 and 30 characters long"
```
spaenleh, line 69: Use interpolation to pass the min and max values.

src/pages/MemberScreen.tsx
```diff
@@ -18,6 +18,7 @@ import { Loader } from '@graasp/ui';
 import EmailPreferenceSwitch from '@/components/main/EmailPreferenceSwitch';
 import LanguageSwitch from '@/components/main/LanguageSwitch';
 import Main from '@/components/main/Main';
+import UsernameForm from '@/components/main/Usernameform';
```
spaenleh, line 21: Be consistent in the import name and file name casing. We prefer `UsernameForm`

spaenleh (changes requested):
Thank you for the changes. 

Please be consistent in your naming conventions. 
Import your components with the same name as the file. 
Use consistent PascalCase for files, components and types. 

Apply the requested changes and request a review.
I have deployed the changes to dev, for the team to test.
Discuss with @MartinaVin to see if she has some improvements to propose design-wise. I am not really a fan of the `Button` for validate and cancel. I think we should use `IconButton`.

Sometimes you use `UserName` sometimes `Username`.

src/components/main/EditingNameField.tsx
```diff
@@ -0,0 +1,68 @@
...
+const verifyUsername = (username: string) => {
+  if (username.trim() === '') {
+    return 'USERNAME_EMPTY_ERROR';
+  }
+  if (username.length < 3 || username.length > 30) {
+    return 'USERNAME_LENGTH_ERROR';
+  }
+  return null;
+};
+let errorMessage;
```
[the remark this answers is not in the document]
mariembencheikh: ok it's noted

src/components/main/EditingNameField.tsx
```diff
@@ -1,45 +1,57 @@
...
-
-    if (errorMessage) {
-      setError(errorMessage ? t(errorMessage) : null);
-    }
+    const errorMessage = verifyUsername(value);
+    setError(
+      errorMessage
+        ? t(errorMessage, {
+            MIN: MIN_USERNAME_LENGTH,
+            MAX: MAX_USERNAME_LENGTH,
```
spaenleh, line 48: Our convention for interpolation keys is to use camelCase.
```suggestion
            min: MIN_USERNAME_LENGTH,
            max: MAX_USERNAME_LENGTH,
```

spaenleh (changes requested):
Nice, nearly there. Also having the tests will be nice !

cypress/e2e/profile.cy.ts
```diff
@@ -1,6 +1,53 @@
...
+  ID_BUTTON_SAVE_USERNAME,
+  ID_BUTTON_SHOW_EDIT_FORM,
+  ID_MEMBER_NAME_TEXT,
+} from '@/config/selectors';
+
+import { BOB, MEMBERS } from '../fixtures/members';
+
+const changeUsername = (newUserName: string) => {
+  cy.setUpApi({ currentMember: BOB });
+  cy.visit('/');
```
spaenleh, line 12: Put this in a `beforeEach` hook.

src/config/selectors.ts
```diff
@@ -15,3 +15,8 @@ export const USER_MENU_SIGN_OUT_OPTION_ID = 'userMenuSignOutOption';
 
 export const DELETE_MEMBER_DIALOG_TITLE_ID = 'alert-dialog-title';
 export const DELETE_MEMBER_DIALOG_DESCRIPTION_ID = 'alert-dialog-description';
+export const ID_BUTTON_SHOW_EDIT_FORM = 'button';
+export const ID_BUTTON_SAVE_USERNAME = 'save-username';
+export const ID_BUTTON_CANCEL_EDITING = 'cancel-editing';
+export const ID_CURRENT_USERNAME = 'current-username';
+export const ID_MEMBER_NAME_TEXT = 'member-name';
```
spaenleh, line 22: Our naming convention is to put the `_ID` as a suffix. And to start by the feature this targets:
`USERNAME_SAVE_BUTTON_ID` for example

package.json
```diff
@@ -88,6 +88,7 @@
     "@testing-library/react": "14.2.2",
     "@testing-library/user-event": "14.5.2",
     "@trivago/prettier-plugin-sort-imports": "4.3.0",
+    "@types/jest": "29.5.12",
```
spaenleh, line 91: Est-ce que tu peux préciser pourquoi tu as ajouté ce packet ? il ne me semble pas que on utilise jest dans ce projet.

cypress/e2e/profile.cy.ts
```diff
@@ -1,6 +1,53 @@
...
+  cy.get(`#${ID_BUTTON_SHOW_EDIT_FORM}`).click();
+  cy.get('input[name=username]').clear();
+  // Find the input field and type the new username
+  cy.get('input[name=username]').type(newUserName);
+};
 describe('template spec', () => {
```
spaenleh, line 18: Change name of test suite.

cypress/e2e/profile.cy.ts
```diff
@@ -1,6 +1,53 @@
...
   it('passes', () => {
```
spaenleh, line 19: Update test name

cypress/e2e/profile.cy.ts
```diff
@@ -1,6 +1,53 @@
...
+    const validUsername = 'validUsername';
+    changeUsername(validUsername);
+    cy.get(`#${ID_BUTTON_SAVE_USERNAME}`).should('not.be.disabled');
+
+    cy.get(`#${ID_BUTTON_SAVE_USERNAME}`).click();
+    cy.wait('@editMember').then(({ request: { body } }) => {
+      expect(body.name).to.equal(validUsername);
+    });
+  });
+  it('button back not modify the username', () => {
```
spaenleh, line 48:
```suggestion
  it('should not update the user name if canceling edit', () => {
```

spaenleh, line 48: Use consistent spacing.

cypress/cypress.d.ts
```diff
@@ -1,6 +1,6 @@
 declare namespace Cypress {
   interface Chainable {
     // todo: improve types
-    setUpApi(args: any): Chainable;
+    setUpApi(args?: any): Chainable;
```
spaenleh, line 4: is this change needed ?

spaenleh (changes requested):

cypress/cypress.d.ts
mariembencheikh: Yes you have right, when i'm working with tests i  was obliged to add "?" and then i forget to remove it

package.json
mariembencheikh: Je ne me souviens pas de ce packet,

cypress/e2e/profile.cy.ts
spaenleh: Please rename this test. `passes` is not a good name.

src/config/selectors.ts
```diff
@@ -15,3 +15,8 @@ export const USER_MENU_SIGN_OUT_OPTION_ID = 'userMenuSignOutOption';
 
 export const DELETE_MEMBER_DIALOG_TITLE_ID = 'alert-dialog-title';
 export const DELETE_MEMBER_DIALOG_DESCRIPTION_ID = 'alert-dialog-description';
+export const BUTTON_SHOW_EDIT_FORM_ID = 'button';
+export const BUTTON_SAVE_USERNAME_ID = 'save-username';
+export const BUTTON_CANCEL_EDITING_ID = 'cancel-editing';
+export const CURRENT_USERNAME_ID = 'current-username';
+export const MEMBER_NAME_TEXT_ID = 'member-name';
```
spaenleh, line 22: I made a suggestion for these, actually I want you to apply the naming suggested. `BUTTON_SHOW_EDIT_FORM_ID` is too broad. Use more specific names. Also they should be consistent with one another. 

Here you have `CURRENT_USERNAME_ID` and `MEMBER_NAME_TEXT_ID`, they both identify the username UI element, one when it is editable, and  the other when it is simply displaying. They should have something In common in the name don't you think ?

Also you actual ids are too generic. `button` is not a good id to use. 

Below are values I think follow these guidelines. Do you agree that they:
- are easier to find and reference
- express better the intent
- are consistent
```suggestion
export const USERNAME_EDIT_BUTTON_ID = 'username-edit-button';
export const USERNAME_SAVE_BUTTON_ID = 'username-save-button';
export const USERNAME_CANCEL_BUTTON_ID = 'username-cancel-button';
export const USERNAME_INPUT_FIELD_ID = 'username-input-field';
export const USERNAME_DISPLAY_ID = 'username-display';
```
Please use these names.

src/components/main/UsernameForm.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <>
+      <Box id={MEMBER_NAME_TEXT_ID}>{member.name}</Box>
+      <IconButton
+        onClick={() => setIsEditing(true)}
+        id={BUTTON_SHOW_EDIT_FORM_ID}
+      >
+        <ModeEdit />
+      </IconButton>
+    </>
```
spaenleh, line 55: I deployed the latest code from the branch to dev and this is the display:
Screenshot 2024-04-12 at 13 23 30
Is it intentional to have the pen icon below the name ?

src/components/main/EditingNameField.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const handleSave = () => {
+    const errorMessage = verifyUsername(newUserName);
+
+    if (!errorMessage) {
+      onSave(newUserName);
+    }
+  };
+  return (
+    <Stack direction="row" alignItems="center" spacing={1}>
+      <TextField
```
spaenleh, line 67: Add the `autofocus` prop to this textfield, so the input is focused when you open it. This will be nicer

spaenleh (changes requested):
Thank you for the changes.
Please fix all comments.

cypress/e2e/profile.cy.ts
```diff
@@ -20,40 +20,50 @@ describe('Change username', () => {
...
-  it('should not update the user name if canceling edit', () => {
+  it('Should not update the user name if canceling edit', () => {
     changeUsername('validUsername');
-    cy.get(`#${BUTTON_CANCEL_EDITING_ID}`).click();
-    cy.get(`#${MEMBER_NAME_TEXT_ID}`).contains(BOB.name);
+    cy.get(`#${USERNAME_CANCEL_BUTTON_ID}`).click();
+    cy.get(`#${USERNAME_DISPLAY_ID}`).contains(BOB.name);
+  });
+
+  it.only('Saves username after trimming trailing space', () => {
```
spaenleh, line 60: Remove `.only`

cypress/e2e/profile.cy.ts
```diff
@@ -1,6 +1,69 @@
...
+    changeUsername('validUsername');
+    cy.get(`#${USERNAME_CANCEL_BUTTON_ID}`).click();
+    cy.get(`#${USERNAME_DISPLAY_ID}`).contains(BOB.name);
+  });
+
+  it.only('Saves username after trimming trailing space', () => {
+    const usernameWithTrailingSpace = 'test  '; // Nom d'utilisateur avec espace à la fin
+    changeUsername(usernameWithTrailingSpace);
+    cy.get(`#${USERNAME_SAVE_BUTTON_ID}`).click();
+    cy.intercept('PUT', '/api/edit-member').as('editMember');
```
spaenleh, line 64: This intercept is wrong. Look at the [api call definition in graasp-query-client](https://github.com/graasp/graasp-query-client/blob/7adee31677b9f0785b3bc4fd5cad9d843dae1ec8/src/api/member.ts#L93)
You test works, because there is already an intercept defined that matches the right call and that is called `editMember` in [the cypress/support/server.ts file](https://github.com/graasp/graasp-account/blob/918ea40f9b1b4fc1f76510c06280c5ab8acf71c8/cypress/support/server.ts#L93)

src/components/main/UsernameForm.tsx
```diff
@@ -0,0 +1,62 @@
...
+      </Grid>
+      <Grid item>
+        <IconButton
+          onClick={() => setIsEditing(true)}
+          id={USERNAME_EDIT_BUTTON_ID}
+        >
+          <ModeEdit />
+        </IconButton>
+      </Grid>
+    </Grid>
```
spaenleh, line 59: In this case `Grid` is not the right choice I think. Prefer [`Stack`](https://mui.com/material-ui/react-stack/) as it allows for 1-dimensional arrangement of data, where grid is 2-dimensional. Also grid has some weird margin/padding behavior. 

```suggestion
    <Stack direction="row" alignItems="center" spacing={1}>
      <Box id={USERNAME_DISPLAY_ID}>{member.name}</Box>
      <IconButton
        onClick={() => setIsEditing(true)}
        id={USERNAME_EDIT_BUTTON_ID}
      >
        <ModeEdit />
      </IconButton>
    </Stack>
```
Also with stack you do not need to wrap the elements again in a `<Grid item ...`

spaenleh (changes requested):

PR #187, opened 2024-04-11
feat: update main component, favicon and font

spaenleh:
In this PR:
- update to use new font
- create a page wrapper that instantiates the Main component from UI, we now have access to the mobile drawer
- update the favicon and the header color

image


src/pages/StorageScreen.tsx
```diff
@@ -13,92 +12,86 @@ import { humanFileSize } from '@/utils/filesize';
...
-              />
-            </Stack>
+    return (
+      <>
+        <Stack direction="row" alignItems="center">
+          <Stack>
+            <div
+              style={{
+                background: theme.palette.primary.main,
+                width: Math.min((current / maximum) * BAR_WIDTH, BAR_WIDTH),
```
ReidyT, line 29: Maybe it could be interesting to export the storage progression in a const to have a better understanding on the calcul and to reuse it in the other div ?

```const storageUsage = (current / maximum) * BAR_WIDTH;```

```suggestion
                width: Math.min(storageUsage, BAR_WIDTH),
```

src/pages/StorageScreen.tsx
```diff
@@ -13,92 +12,86 @@ import { humanFileSize } from '@/utils/filesize';
...
+                height: BAR_HEIGHT,
+              }}
+            />
           </Stack>
           <Stack>
-            {humanFileSize(current)}/{humanFileSize(maximum)}
+            <div
+              style={{
+                background: 'lightgrey',
+                width: BAR_WIDTH - (current / maximum) * BAR_WIDTH,
```
ReidyT, line 38:
```suggestion
                width: BAR_WIDTH - storageUsage,
```

ReidyT (approved):
Thank you for the PR ! I just let a small comment but it looks good for me 🙌

spaenleh:
@ReidyT Thank you for the suggestion, I went a bit further and used the `<progress>` HTML component. Here is how it looks:
Webkit/Blink (Chrome, Safari)
Screenshot 2024-04-12 at 16 29 38

Firefox:
Screenshot 2024-04-12 at 16 29 48

The progress is not as easy to style on firefox, but I tried to make it match as best as possible.

PR #244, opened 2024-05-30
feat: create read only profile page

mariembencheikh:
#241


src/pages/MemberScreen.tsx
```diff
@@ -1,158 +1,19 @@
...
-import notifier from '@/config/notifier';
-import { hooks, mutations } from '@/config/queryClient';
-import {
-  MEMBER_PROFILE_ANALYTICS_SWITCH_ID,
-  MEMBER_PROFILE_LANGUAGE_SWITCH_ID,
-} from '@/config/selectors';
-import { COPY_MEMBER_ID_TO_CLIPBOARD } from '@/types/clipboard';
-import { copyToClipboard } from '@/utils/clipboard';
 
 const MemberProfileScreen = (): JSX.Element | null => {
```
pyphilia, line 7: Rename file like the name of the component `MemberProfileScreen`

cypress/e2e/memberProfile.cy.ts
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  it('displays the correct member info', () => {
+    // displays the correct member name
+    cy.get(`#${USERNAME_DISPLAY_ID}`).should('contain', BOB.name);
+    // displays the correct member email
+    cy.get(`#${MEMBER_PROFILE_EMAIL_ID}`).should('contain', BOB.email);
+  });
+
+  it('displays the correct public profile info', () => {
```
pyphilia, line 31: Let's also add a test when the public profile is empty. It's interesting to check, because sometimes when info is unavailable a page can crash.

src/components/main/MemberPersonalInformation.tsx
```diff
@@ -0,0 +1,79 @@
...
+          <Typography variant="h5">
+            {t('PERSONAL_INFORMATION_TITLE')}
+          </Typography>
+        </Grid>
+        <Grid item xs={12} sm={6} container justifyContent="flex-end">
+          <Button
+            variant="contained"
+            color="primary"
+            onClick={() => goTo(EDIT_MEMBER_INFO)}
+          >
```
pyphilia, line 43: Use `react-router-dom`'s `Link` to avoid using `navigate` directly. https://reactrouter.com/de/main/components/link

pyphilia (changes requested):
Great PR! I added my comments, as well as general/design comments here:

I think it was not especially wanted to have this break effect (title-button). I think it's okay to not break, but rather use `xs=10` `xs=2` or 11/1
Screenshot 2024-05-31 at 13 19 19

I think the display is bit strange when the public profile is empty. Maybe let's have some indication that some fields are empty?
Screenshot 2024-05-31 at 13 23 45

Typo dans "membre depuis"
Screenshot 2024-05-31 at 13 28 03

Also I know it was not in the mockup, but I think it would be interesting to have the link to the public profile in the read page as well.

src/components/main/MemberPersonalInformation.tsx
```diff
@@ -0,0 +1,72 @@
...
+          </Typography>
+        </Grid>
+        <Grid item xs={6} sm={6} container justifyContent="flex-end">
+          <Link to={EDIT_MEMBER_INFO} className="link">
+            <Button variant="contained" color="primary">
+              {t('EDIT_BUTTON')}
+            </Button>
+          </Link>
+        </Grid>
+      </Grid>
```
spaenleh, line 40: I do not see the advantage of using a `Grid` here. A simple stack is less verbose and more flexible (pun intended).
```suggestion
      <Stack direction="row" justifyContent="space-between" spacing={2}>
        <Typography variant="h5">{t('PERSONAL_INFORMATION_TITLE')}</Typography>
        <Link to={EDIT_MEMBER_INFO} className="link">
          <Button variant="contained" color="primary">
            {t('EDIT_BUTTON')}
          </Button>
        </Link>
      </Stack>
```

With grid since you divide the space into 2 equal zones (`sx={6}`) you can have un-natural wrapping:
Screenshot 2024-06-03 at 08 58 23
I would expect the text to have the space to not have to wrap on a second line here. 

When using the provided Stack solution it does not happen, since the text container is not constrained to taking only half the width:
Screenshot 2024-06-03 at 09 00 57
I also think the text and button should be aligned to the top (default, otherwise the space between the text top and the container is not the same as on the side. Bellow is what I think looks best:
Screenshot 2024-06-03 at 09 05 42

src/components/main/MemberPersonalInformation.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <Grid container spacing={2}>
```
spaenleh, line 41: If using the same spacing, we loose a hierarchical sense when on smaller screens. I think label and setting should be closer together and there should be a larger space between the groups. 
Screenshot 2024-06-03 at 09 07 01

src/components/main/MemberPersonalInformation.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <Grid item xs={12} sm={4}>
+          <Typography variant="body1" color="textSecondary">
+            {t('PROFILE_MEMBER_NAME')}
+          </Typography>
+        </Grid>
+        <Grid item xs={12} sm={8}>
+          <Typography variant="body1" id={USERNAME_DISPLAY_ID}>
+            {member?.name}
+          </Typography>
+        </Grid>
```
spaenleh, line 51: This is very repetitive, you should create a component that abstracts this layout.

src/components/main/MemberPublicProfile.tsx
```diff
@@ -0,0 +1,124 @@
+import { Link } from 'react-router-dom';
+
+import FacebookIcon from '@mui/icons-material/Facebook';
+import LinkedInIcon from '@mui/icons-material/LinkedIn';
+import TwitterIcon from '@mui/icons-material/Twitter';
```
spaenleh, line 5: We are transitioning to lucide.dev icons. Can you use them instead ?

src/components/main/MemberPublicProfile.tsx
```diff
@@ -0,0 +1,124 @@
...
+              variant="body1"
+              color="textSecondary"
+              id={PUBLIC_PROFILE_BIO_ID}
+            >
+              {t('PUBLIC_PROFILE_BIO_EMPTY_MSG')}
+            </Typography>
+          )}
+        </Box>
+
+        <Stack direction="row" spacing={2}>
```
spaenleh, line 63: Same comment about re-usable component here.

src/components/main/MemberPublicProfile.tsx
```diff
@@ -0,0 +1,124 @@
...
+          <Typography variant="body1" color="textSecondary">
+            <LinkedInIcon />
+          </Typography>
+          {data?.linkedinID ? (
+            <Typography variant="body1" id={PUBLIC_PROFILE_LINKEDIN_ID}>
+              {data?.linkedinID}
```
spaenleh, line 69: I see that currently you display the social network ids as-is. This is what we store, but I am not sure this is what makes the most sens for the user. they are not clickable, they can not verify that it links to the right info. Also it is another way of displaying them. We should have the same way to display them here and in the library member page. We should discuss this in the next meeting.

spaenleh (changes requested):
Good work. I have a few comments:

- [x] change PR title to conform to the standard (for reference see [the style guide](https://graasp.github.io/docs/developer/style-guide/#conventional-pr-title))
- [ ] your PR has not been rebased, it is not dramatic, but would be better to do in the future, as it provides a clearer view of the changes you want to apply
- [x] the Changelog has been formatted, you should not format files you do not care about. To avoid committing files you do not wish to change, review your changes locally, or in github and revert any change you do not wish to make. (Refer to [the style guide](https://graasp.github.io/docs/developer/style-guide/#review-your-changes-before-asking-for-reviews) for more info)
- [x] Once you edit one part of the profile, there is no way to return back to the profile page. Add a button even if only temporary, so that users are not lost.
- [x] In general you have repeating code, create components to re-use and make it easier to change layout more uniformly.

src/components/main/MemberPersonalInformation.tsx
```diff
@@ -0,0 +1,47 @@
...
+          title={t('PROFILE_MEMBER_NAME')}
+          content={member?.name}
+          contentId={USERNAME_DISPLAY_ID}
+        />
+        <MemberProfileItem
+          title={t('PROFILE_EMAIL_TITLE')}
+          content={member?.email}
+          contentId={MEMBER_PROFILE_EMAIL_ID}
+        />
+        <MemberProfileItem title={t('PASSWORD_TITLE')} contentId="" />
```
spaenleh, line 41: Specify a value for `contentId`.

src/pages/EditPublicProfileScreen.tsx
```diff
@@ -220,27 +229,31 @@ const PublicProfileScreen = (): JSX.Element => {
             }
             label={t('PUBLIC_PROFILE_VISIBILITY')}
           />
+          <Stack direction="row" spacing={2}>
+            <Link to={PROFILE_PATH}>
+              <Button variant="outlined">{t('CANCEL_BUTTON')}</Button>
```
spaenleh, line 234: I don't think it is logical to have the button say "Cancel" ... "Close" would be more appropriate.

src/pages/MemberProfileScreen.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { useAccountTranslation } from '@/config/i18n';
+
+const MemberProfileScreen = (): JSX.Element | null => {
+  const { t } = useAccountTranslation();
+  return (
+    <Stack spacing={2}>
+      <Typography variant="h4" component="h1">
+        {t('PROFILE_TITLE')}
+      </Typography>
+      <Divider sx={{ my: 2 }} />
```
spaenleh, line 14: The `my` has no effect on this component (at least from what I can see)
```suggestion
      <Divider />
```

src/pages/MemberProfileScreen.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import MemberPersonalInformation from '@/components/main/MemberPersonalInformation';
+import MemberPublicProfile from '@/components/main/MemberPublicProfile';
+import { useAccountTranslation } from '@/config/i18n';
+
+const MemberProfileScreen = (): JSX.Element | null => {
+  const { t } = useAccountTranslation();
+  return (
+    <Stack spacing={2}>
+      <Typography variant="h4" component="h1">
```
spaenleh, line 11: I like that you used `component` and `variant` but I think we should go for instead:
```suggestion
      <Typography variant="h2" component="h1">
```

src/pages/MemberProfileScreen.tsx
```diff
@@ -0,0 +1,21 @@
+import { Divider, Stack, Typography } from '@mui/material';
+
+import MemberPersonalInformation from '@/components/main/MemberPersonalInformation';
+import MemberPublicProfile from '@/components/main/MemberPublicProfile';
+import { useAccountTranslation } from '@/config/i18n';
+
+const MemberProfileScreen = (): JSX.Element | null => {
```
spaenleh, line 7: The return type is too wide, there is no way this returns `null`.

src/components/main/MemberPublicProfileItem.tsx
spaenleh, line 1: I appreciate the effort on this component, but maybe it tries to do a bit too much things ? Maybe let it only display the links, and display the bio differently. Also change the name to be less linked to the Public Profile. Reflect more what the component is able to do instead of where it is used. This should improve re-usability. Continue like this !

src/components/main/MemberPublicProfile.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <MemberPublicProfileItem
+        title={t('PUBLIC_PROFILE_BIO')}
+        content={bio}
+        emptyMessage={t('PUBLIC_PROFILE_BIO_EMPTY_MSG')}
+        contentId={PUBLIC_PROFILE_BIO_ID}
+        href=""
+      />
+      <MemberPublicProfileItem
+        icon={<Linkedin fill="grey" strokeWidth={0} />}
+        content={linkedinID ? socialLinks.sanitize('linkedin', linkedinID) : ''}
```
spaenleh, line 49: As much as I like the clever rendering you do in the component, I do not really like that you have to pass the sanitized value twice. Either supply a human readable text for the link content, or use just one prop for the href and display it like that.

src/components/main/MemberProfileItem.tsx
```diff
@@ -0,0 +1,27 @@
...
+      <Typography variant="body1" color="textSecondary">
+        {title}
+      </Typography>
+    </Grid>
+    <Grid item xs={12} sm={9}>
+      <Typography variant="body1" id={contentId}>
+        {content}
+      </Typography>
+    </Grid>
+  </>
```
spaenleh, line 24: Note: I checked and it looks like we would not really need to change the display direction from row to column on small screen sizes.

src/components/main/MemberPersonalInformation.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+const MemberPersonalInformation = (): JSX.Element => {
+  const { data: member } = hooks.useCurrentMember();
+  const { t } = useAccountTranslation();
+
+  return (
+    <RoundedStack>
+      <Stack direction="row" justifyContent="space-between">
+        <Typography variant="h5">{t('PERSONAL_INFORMATION_TITLE')}</Typography>
+        <Link to={EDIT_MEMBER_INFO} className="link">
```
spaenleh, line 24: Not sure why you specified the `link` className ?

src/components/main/MemberPublicProfile.tsx
```diff
@@ -0,0 +1,75 @@
...
+  PUBLIC_PROFILE_BIO_ID,
+  PUBLIC_PROFILE_FACEBOOK_ID,
+  PUBLIC_PROFILE_LINKEDIN_ID,
+  PUBLIC_PROFILE_TWITTER_ID,
+} from '@/config/selectors';
+
+import MemberPublicProfileItem from './MemberPublicProfileItem';
+import RoundedStack from './RoundedStack';
+
+const MemberPublicProfile = (): JSX.Element | null => {
```
spaenleh, line 21: Return type can not be `null`.

spaenleh (changes requested):
Thank you for the changes, I think I will stop bothering you. But I think you should really perform an auto-review to evaluate your code and detect potential refactoring to improve the readability and elegance of the code. This is a good exercise. If you do not find it beautiful, there is maybe a better way to do it. Just something to think about !

spaenleh:
@mariembencheikh I think you should add a container inside the profile page. Recommended max size would be 1000px.

spaenleh (approved):
Great work ! 🚀

src/pages/MemberProfileScreen.tsx
```diff
@@ -0,0 +1,23 @@
+import { Container, Divider, Stack, Typography } from '@mui/material';
+
+import MemberPersonalInformation from '@/components/main/MemberPersonalInformation';
+import MemberPublicProfile from '@/components/main/MemberPublicProfile';
+import { useAccountTranslation } from '@/config/i18n';
+
+const MemberProfileScreen = (): JSX.Element => {
+  const { t } = useAccountTranslation();
+  return (
+    <Container sx={{ maxWidth: '1000px' }}>
```
pyphilia, line 10: No need, Container has by default a maxWidth of 1200.
```suggestion
    <Container>
```

pyphilia (changes requested):
Thanks for all the changes, I suggest other minor changes, we can merge as soon as you apply them. Thanks 🦆

src/components/main/MemberPublicProfile.tsx
```diff
@@ -0,0 +1,94 @@
...
+          contentId={PUBLIC_PROFILE_LINKEDIN_ID}
+          content={t('PUBLIC_PROFILE_LINKEDIN_EMPTY_MSG')}
+        />
+      )}
+
+      {twitterID ? (
+        <DisplayingMemberPublicProfileLinks
+          icon={<TwitterIcon />}
+          contentId={PUBLIC_PROFILE_TWITTER_ID}
+          href={socialLinks.sanitize('linkedin', twitterID)}
```
pyphilia, line 66:
```suggestion
          href={socialLinks.sanitize('twitter', twitterID)}
```

src/components/main/MemberPublicProfile.tsx
```diff
@@ -0,0 +1,94 @@
...
+          icon={<TwitterIcon />}
+          contentId={PUBLIC_PROFILE_TWITTER_ID}
+          content={t('PUBLIC_PROFILE_TWITTER_EMPTY_MSG')}
+        />
+      )}
+      {facebookID ? (
+        <DisplayingMemberPublicProfileLinks
+          icon={<FacebookIcon />}
+          contentId={PUBLIC_PROFILE_FACEBOOK_ID}
+          href={socialLinks.sanitize('linkedin', facebookID)}
```
pyphilia, line 80:
```suggestion
          href={socialLinks.sanitize('facebook', facebookID)}
```

pyphilia (changes requested):
I noticed a last bug, please write a test to also verify that the links are correct.

pyphilia (approved):

PR #247, opened 2024-06-05
fix: fix avatar size constant

### LINKED ISSUE #246: Fix avatar size

mariembencheikh:
fix #246

### DIFF
--- a/src/components/main/AvatarUploader.tsx
+++ b/src/components/main/AvatarUploader.tsx
@@ -4,6 +4,7 @@ import { Dialog, Stack, styled, useTheme } from '@mui/material';
 
 import { ImageUp as ImageUpIcon } from 'lucide-react';
 
+import { AVATAR_SIZE } from '@/config/constants';
 import { useAccountTranslation } from '@/config/i18n';
 import {
   AVATAR_UPLOAD_ICON_CY,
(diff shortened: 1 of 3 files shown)


spaenleh (changes requested):
Thank you for the quick fix, it solves the issue.

However I think this only fixes the surface of the problem and the core can still be subject to breakage when changing/refactoring the code. Let me explain:
- There are 3 constants in the app that somehow relate to sizing the big profile picture:
  - AVATAR_SIZE (previously 256, now 200)
  - THUMBNAIL_SETTING_MAX_HEIGHT/WIDTH (200)
  - THUMBNAIL_DIMENSION (200)
- The `AvatarUploader` component allows to set `avatarSize` via an optional prop. This component is used only once, and is passed a prop value for `avatarSize` coming from the AVATAR_SIZE constant, so the THUMBNAIL_DIMENSION is actually never used.
- Different parts of the AvatarUploader UI are sized with different constants, it works if they have the same value, but breaks if one of them is changed. 

In the light of these elements can you reconsider the fix and make it less fragile to changes. I think reducing the number or constants that affect the same values should be considered as well as ensuring that the elements that should be the same size depend on the same variable instead of different ones that hold the same value. 

Let me know if something is not clear. Also please take this as a nitpicking comment from my side, that tries to improve the current code, instead of duck taping it 😄

spaenleh (approved):
Nice !

PR #261, opened 2024-06-14
feat: add preferences to settings page

mariembencheikh:
#260


spaenleh (changes requested):
Nice work ! 

I have left some comments. And there is an inconsistency in the layout when you switch between read and edit in the settings: 

https://github.com/graasp/graasp-account/assets/39373170/7b468130-9e53-4cf1-bc68-db6038b040ff

cypress/e2e/editSettings.cy.ts
```diff
@@ -13,22 +14,28 @@ import { buildDataCySelector } from '../support/utils';
 
 const currentMember = CURRENT_MEMBER;
 
-const expectEnableSaveActionsInRequestToBe = (enableSaveActions: boolean) =>
-  cy
-    .wait('@editMember')
+const expectEnableSaveActionsInRequestToBe = (enableSaveActions: boolean) => {
+  cy.get(`#${PREFERENCES_SAVE_BUTTON_ID}`).click();
```
spaenleh, line 18: I would not put this inside the expect function as it is not related. 

Close the editing view, and then assert that the request had the expected values.

cypress/e2e/editSettings.cy.ts
```diff
@@ -13,22 +14,28 @@ import { buildDataCySelector } from '../support/utils';
...
-    .get(buildDataCySelector(MEMBER_PROFILE_ANALYTICS_SWITCH_ID, 'input'))
-    .should(`${enabled ? '' : 'not.'}be.checked`);
+const expectAnalyticsSwitchToBe = (enabled: boolean) => {
+  cy.visit(MANAGE_ACCOUNT_PATH);
+  cy.get(`#${MEMBER_PROFILE_EDIT_PREFERENCES_BUTTON_ID}`).click();
+  cy.wait('@getCurrentMember');
+
+  cy.get(
+    buildDataCySelector(MEMBER_PROFILE_ANALYTICS_SWITCH_ID, 'input'),
+  ).should(`${enabled ? '' : 'not.'}be.checked`);
```
spaenleh, line 37: use an if-else structure instead of the ternary for more clarity.

src/components/main/EditMemberPreferences.tsx
```diff
@@ -32,11 +27,13 @@ import {
   PREFERENCES_SAVE_BUTTON_ID,
 } from '@/config/selectors';
 
-import RoundedStack from '../common/RoundedStack';
-
-const EditMemberPreferences = (): JSX.Element | null => {
+type EditPreferencesProp = {
+  onCancel: () => void;
```
spaenleh, line 31: Don't call it `onCancel` if you invoke it also after saving the settings. Use `onClose` which is more neutral.

src/components/main/MemberPreferences.tsx
```diff
@@ -25,50 +24,58 @@ const MemberPreferences = (): JSX.Element => {
   const languageName = langs[languageCode];
 
   const { t } = useAccountTranslation();
+  const [isEditing, setIsEditing] = useState(false);
+  const handleEditClick = () => {
+    setIsEditing(true);
+  };
+
+  const handleCancelClick = () => {
```
spaenleh, line 32: Same here, I think it should be called: `handleClose`

spaenleh (approved):
Looks good, just a few nitpicks about naming.

src/components/main/EditMemberPreferences.tsx
```diff
@@ -27,12 +27,12 @@ import {
   PREFERENCES_SAVE_BUTTON_ID,
 } from '@/config/selectors';
 
-type EditPreferencesProp = {
-  onCancel: () => void;
+type oncloseProp = {
```
spaenleh, line 30: This should not have been changed. 
```suggestion
type EditPreferencesProp = {
```

spaenleh (commented):
Nice ! Just one more thing 😄

spaenleh (approved):

PR #263, opened 2024-06-17
fix: issue on delete-account-button

spaenleh:
new-delete-interface

### DIFF
--- a/src/components/main/DeleteMemberDialogContent.tsx
+++ b/src/components/main/DeleteMemberDialogContent.tsx
@@ -31,6 +31,10 @@ const DeleteMemberDialogContent = ({ id, closeModal }: Props): JSX.Element => {
     'DELETE_CONFIRMATION_VALUE',
   );
 
+  // confirmation is disabled when the two text do not match
+  const isConfirmationDisabled =
+    confirmationDeleteValue !== confirmationDeleteTextToCompare;
+
   return (
     <>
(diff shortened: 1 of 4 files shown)


src/components/main/DeleteMemberDialogContent.tsx
```diff
@@ -31,6 +31,10 @@ const DeleteMemberDialogContent = ({ id, closeModal }: Props): JSX.Element => {
     'DELETE_CONFIRMATION_VALUE',
   );
 
+  // confirmation is disabled when the two text do not match
```
pyphilia, line 34:
```suggestion
  // confirmation is disabled when the two texts do not match
```

src/components/main/DeleteMemberDialogContent.tsx
```diff
@@ -50,8 +54,9 @@ const DeleteMemberDialogContent = ({ id, closeModal }: Props): JSX.Element => {
         <TextField
           value={confirmationDeleteValue}
           fullWidth
+          required
           variant="outlined"
-          placeholder={confirmationDeleteTextToCompare}
+          // placeholder={confirmationDeleteTextToCompare}
```
pyphilia, line 59: why remove the placeholder?

pyphilia (approved):

src/components/main/DeleteMemberDialogContent.tsx
spaenleh: It seems like it was confusing for some users to have the text already written in the field.

PR #283, opened 2024-07-04
feat: update and test the edit personal information interface

mariembencheikh:
#257


spaenleh:
@mariembencheikh It looks like in your last commit you formatted more files than necessary with prettier.

The issue looked like it was only on `src/components/main/MainMenu.tsx`. But you formatted also the yarn config, changelog and public/.siteconfig. 

What command did you use ? `yarn prettier:write` should not touch these files. Did you use a command in VSCode ? 

Can you please revert your last commit, and only format `src/components/main/MainMenu.tsx` ? Thank you.

spaenleh:
@mariembencheikh If this PR is based on code from a previous branch with a PR still open please let us know, so we don't review code twice. You can for example say "This PR is based on (number of the other PR), please review it before this one."

pyphilia (changes requested):
Thanks for your work! Here are my additional general comments:

- compared to your figma design, values of the properties should be aligned more on the right. Could we have this same design? I think splitting the container with Grid can make the job.
Screenshot 2024-07-05 at 13 45 54
- you shouldn't be able to save if nothing changed.
- could we have the close and save buttons below the form (instead of top)?
- can we change "Password" (not the title, but below) by "Update your password by clicking on the button Edit"?
Screenshot 2024-07-05 at 14 02 03

src/components/main/EditMemberPersonalInformation.tsx
```diff
@@ -0,0 +1,129 @@
...
+    return 'USERNAME_LENGTH_ERROR';
+  }
+  return null;
+};
+type EditMemberPersonalInformationProp = {
+  member: {
+    id: string;
+    name: string;
+    email: string;
+  };
```
pyphilia, line 35:
```suggestion
  member: Member
```

src/components/main/EditMemberPersonalInformation.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  // save changes
+  const handleSave = () => {
+    const errorMessage = verifyUsername(newUserName ?? '');
+
+    if (!errorMessage) {
+      if (member) {
+        editMember({
+          id: member?.id,
```
pyphilia, line 68:
```suggestion
          id: member.id,
```

src/components/main/EditMemberPersonalInformation.tsx
```diff
@@ -0,0 +1,129 @@
...
+          name: newUserName?.trim(),
```
pyphilia, line 69:
```suggestion
          name: newUserName.trim(),
```

pyphilia (changes requested):
Thanks for the quick changes, some issues:
- on mobile it goes out of the field, probably we prefer a break at `sm` instead of `xs`
- I just noticed but the way to edit profile and settings is different. One uses fullwidth fields and the other one keeps the titles but shows selects. I think I prefer the second solution (titles + fields) especially if we want to add more info next to the fields it will look nicer I think. 
- editing settings does not match the same column size (it's more like 30% 70% unlike the read mode). I provide you a screenshot of mobile, here I expect break lines. 
- The save and close buttons are still at the top? Would be better below, especially for small screens.
- save buttons are still enabled even if I didn't change anything
- and could you use dense buttons for edit/close/save? they are quite big I think.

Screenshot 2024-07-05 at 15 40 46


Screenshot 2024-07-05 at 15 36 53

src/components/main/MemberPersonalInformation.tsx
```diff
@@ -1,46 +1,72 @@
...
+  const handleEditClick = () => {
+    setIsEditing(true);
+  };
+
+  const handleClose = () => {
+    setIsEditing(false);
+  };
+  if (member) {
+    return (
+      <Stack id={EDIT_MEMBER_INFO_FORM_ID}>
```
spaenleh, line 33: Mhmm not sure you want to put this id here... it will not change when you edit the member information. Was this intended ?

src/components/main/MemberProfileItem.tsx
```diff
@@ -12,14 +12,25 @@ const MemberProfileItem = ({
...
-  <Stack direction="row" spacing={2} alignItems="center">
-    <Typography variant="body1" color="textSecondary">
-      {title}
-    </Typography>
-    <Typography variant="body1" id={contentId}>
-      {content}
-    </Typography>
-  </Stack>
+  <Grid container alignItems="center">
+    <Grid item xs={6}>
```
spaenleh, line 16: Here you do not use the same breakpoints as when you edit the values.

src/components/main/MemberProfileItem.tsx
```diff
@@ -12,14 +12,25 @@ const MemberProfileItem = ({
...
+      </Typography>
+    </Grid>
+    <Grid item xs={6}>
+      <Typography
+        variant="body1"
+        id={contentId}
+        style={{
+          overflow: 'hidden',
+          textOverflow: 'ellipsis',
+        }}
```
spaenleh, line 28: Don't use `style`. Use `sx` prop, but first check if there are not native TextfieldProps that can be used.

src/langs/en.json
```diff
@@ -42,6 +42,7 @@
   "PROFILE_AVATAR_INFORMATION": "Update Profile picture",
   "PROFILE_AVATAR_CURRENT_ALT": "Profile Picture",
   "PASSWORD_TITLE": "Password",
+  "PASSWORD_SETTINGS_INFORMATION": "Update your password by clicking on the button Edit",
```
spaenleh, line 45:
```suggestion
  "PASSWORD_SETTINGS_INFORMATION": "Update your password by clicking on the 'Edit' button",
```

src/langs/fr.json
```diff
@@ -42,6 +42,7 @@
   "PROFILE_AVATAR_INFORMATION": "Changer l'image de profil",
   "PROFILE_AVATAR_CURRENT_ALT": "Image de profil",
   "PASSWORD_TITLE": "Mot de passe",
+  "PASSWORD_SETTINGS_INFORMATION": "Changer le mot de passe en cliquant sur le bouton Modifier",
```
spaenleh, line 45:
```suggestion
  "PASSWORD_SETTINGS_INFORMATION": "Changez votre mot de passe en cliquant sur le bouton 'Modifier'",
```

src/langs/fr.json
```diff
@@ -86,5 +87,6 @@
   "SECURITY_TIPS_INFORMATION": "Vous n'avez pas défini de mot de passe, vous pouvez en définir un pour accéder à votre compte sans ouvrir votre email.",
   "PERSONALISATION_TITLE": "Personnalisation",
   "PERSONALISATION_INFORMATION": "Laissez votre personnalité et votre unicité briller en ajoutant une photo de profil !",
-  "SAVE_CHANGES_TEXT": "Enregistrer modifications"
+  "SAVE_CHANGES_TEXT": "Enregistrer modifications",
```
spaenleh, line 90:
```suggestion
  "SAVE_CHANGES_TEXT": "Enregistrer les modifications",
```

src/pages/PasswordSettingsScreen.tsx
```diff
@@ -8,8 +8,16 @@ import { FAILURE_MESSAGES } from '@graasp/translations';
...
+import {
+  CONFIRM_PASSWORD_ID,
+  NEW_PASSWORD_ID,
+  PASSWORD_SAVE_BUTTON_ID,
+} from '@/config/selectors';
+
+type onCloseProp = {
+  onClose: () => void;
+};
+const PasswordSettings = ({ onClose }: onCloseProp): JSX.Element => {
```
spaenleh, line 20: The name of the file does not mirror the name of the component.

src/pages/PasswordSettingsScreen.tsx
```diff
@@ -8,8 +8,16 @@ import { FAILURE_MESSAGES } from '@graasp/translations';
...
 import { mutations } from '@/config/queryClient';
-
-const PasswordSettings = (): JSX.Element => {
+import {
+  CONFIRM_PASSWORD_ID,
+  NEW_PASSWORD_ID,
+  PASSWORD_SAVE_BUTTON_ID,
+} from '@/config/selectors';
+
+type onCloseProp = {
```
spaenleh, line 17: Why is this type named like this ?

src/pages/PasswordSettingsScreen.tsx
```diff
@@ -20,6 +28,7 @@ const PasswordSettings = (): JSX.Element => {
     string | null
   >();
   const { mutate: updatePassword } = mutations.useUpdatePassword();
+  const [isEditing, setIsEditing] = useState(false);
```
spaenleh, line 31: Same comment as for the other states, use a better name, this tracks whether there are changes, not whether you are showing the read or write interface ...

src/pages/PasswordSettingsScreen.tsx
```diff
@@ -34,10 +43,11 @@ const PasswordSettings = (): JSX.Element => {
     return newPasswordIsNotEmpty || confirmPasswordIsNotEmpty;
   };
 
-  const onClose = () => {
+  const onCancel = () => {
     setCurrentPassword('');
     setNewPassword('');
     setConfirmPassword('');
+    setIsEditing(false);
```
spaenleh, line 50: I don't think you need to revert the state to the default as the component is destroyed.

src/pages/PasswordSettingsScreen.tsx
```diff
@@ -55,39 +65,42 @@ const PasswordSettings = (): JSX.Element => {
...
     onClose();
   };
 
   const handleCurrentPasswordInput = (event: ChangeEvent<HTMLInputElement>) => {
     setCurrentPassword(event.target.value);
+    setIsEditing(true);
   };
+
   const handleNewPasswordInput = (event: ChangeEvent<HTMLInputElement>) => {
     setNewPassword(event.target.value);
```
spaenleh, line 84: Extract event.target.value to a constant locally.

src/pages/PasswordSettingsScreen.tsx
```diff
@@ -55,39 +65,42 @@ const PasswordSettings = (): JSX.Element => {
...
+    if (!event.target.value) {
+      setNewPasswordError(FAILURE_MESSAGES.PASSWORD_EMPTY_ERROR);
+    } else if (!isPasswordStrong(event.target.value)) {
+      setNewPasswordError(FAILURE_MESSAGES.PASSWORD_WEAK_ERROR);
+    } else {
+      setNewPasswordError(null);
+    }
   };
   const handleConfirmPasswordInput = (event: ChangeEvent<HTMLInputElement>) => {
     setConfirmPassword(event.target.value);
```
spaenleh, line 95: Same here

spaenleh (changes requested):
Thank you for the work. Here are my comments let me know if you have questions.

src/components/main/EditMemberPersonalInformation.tsx
```diff
@@ -0,0 +1,150 @@
...
+    onClose();
+  };
+
+  return (
+    <RoundedStack>
+      <Typography variant="h5" component="h1">
+        {t('PERSONAL_INFORMATION_TITLE')}
+      </Typography>
+
+      <Grid container alignItems="center">
```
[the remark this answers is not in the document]
mariembencheikh: Here, i applied @pyphilia's request review, she asked me to splitting the container with Grid compared to my figma design.

src/components/main/MemberPersonalInformation.tsx
mariembencheikh: it's just to check in cypress the display of the edit form

src/components/main/EditMemberPreferences.tsx
```diff
@@ -36,24 +36,48 @@ const EditMemberPreferences = ({
...
-      const cheked = event.target.checked;
-      setSwitchedSaveActions(cheked);
-      setIsEditing(true);
+      const valueChecked = event.target.checked;
+      setSwitchedSaveActions(valueChecked);
+      setHasModifications(
+        valueChecked !== initialSaveActions ||
+          selectedLang !== initialLang ||
+          selectedEmailFreq !== initialEmailFreq,
+      );
```
spaenleh, line 63: Since you have to take into account all the values, I really don't think that having it in a state saves you anything. You could just simply compute it in a const.

src/components/main/EditPassword.tsx
```diff
@@ -76,25 +75,28 @@ const PasswordSettings = ({ onClose }: onCloseProp): JSX.Element => {
   };
 
   const handleCurrentPasswordInput = (event: ChangeEvent<HTMLInputElement>) => {
-    setCurrentPassword(event.target.value);
-    setIsEditing(true);
+    const passwordTarget = event.target.value;
+    setCurrentPassword(passwordTarget);
+    setHasModifications(true);
```
spaenleh, line 80: Is it correct that you did not fix the issue where the button never goes back to gray If I undo all the changes ?

spaenleh (changes requested):
Nice ! Just a few things :)

PR #291, opened 2024-07-18
feat: create a specification for the get all uploaded files

mariembencheikh:


specificationDocument.md
```diff
@@ -0,0 +1,131 @@
...
+    - name: Must be a string representing the name of the file.
+    - size: Must be a number representing the size of the file in bytes.
+    - createdAt: must be a string representing the date and time.
+    - path: must be a string representing the path indicating the location of the file.
+    - parent.id: must be a string representing the ID of the parent folder (if available).
+    - parent.name: must be a string representing the name of the parent folder (if available).
+- What happens on failure?
+    - Possible Causes of Failure:
+        - Authentication Failure: User is not authenticated or token is invalid.
+        - Permission Denied: User does not have the required permissions to access the item.
```
spaenleh, line 65: We are not asking for a specific item, so how could we not have the required permission to access the item ? The item is either included in the response because we can access it, or not included because we can't access it.

specificationDocument.md
```diff
@@ -0,0 +1,131 @@
...
+        - Folder Not Found: The specified folder ID (parent ID) does not exist.
```
spaenleh, line 66: I guess this could happen when we query the parent item and it somehow does not exist? In that case, does it fail the whole request, or only the one file that we can't get the parent of ?

specificationDocument.md
```diff
@@ -0,0 +1,131 @@
...
+        }
+        ```
+
+
+### Frontend design
+
+- Is this a hook or a mutation? 
+  - This is a hook because we are fetching data from the backend to display in the frontend UI. 
+- What are the parameters of the hook?
+  -  None, as we are fetching a list of files from items in general.
```
spaenleh, line 108: See my comment about pagination.

specificationDocument.md
```diff
@@ -0,0 +1,131 @@
...
+        name: string
+        size: number
+        createdAt: datetime
+        path: string 
+        parentId: string (foreign key reference to the parent folder, if applicable)
+
+        For folders:
+        id: string
+        name: string
+       ```
```
spaenleh, line 131: Thank you for trying to fill this part, I think we will add a link to the specifications file in the backend repo once it exists. It looks mostly accurate though. Good job !

spaenleh (commented):
🎆 
Thank you for writing this specification ! I have added my comments. I let others chime in and we should discuss this live at the end of the next standup to finalize it. 

About the file placement in the project: either inside the folder for the storage components, or in a dedicated `/specs` or `/docs` folder at the root. The file should have the name of the feature.

specificationDocument.md
mariembencheikh: Yes, only the one file that we can't get the parent of

spaenleh: So, is it considered a cause of failure for the whole request ? Or does it simply not return the file whos parent we can't get, but the rest of the files are returned so the request is successful ?

docs/DisplayUploadedFilesInStorage.md
```diff
@@ -0,0 +1,171 @@
...
+        }
+        ```
+    - 500 Internal Server Error
+        ``` json
+        {
+            "status": "error",
+            "error_code": "500",
+            "message": "An unexpected error occurred. Please try again later."
+        }
+        ```
```
ReidyT, line 136: Maybe I would reorganise a bit this part like the following. You could even include the successful path. What do you think ?



- Responses:
  - 200 OK
    - The request was successful.
    - schema:
    ```json
    {
        "data": [
          {
            "id":"string", // Item's id
            "name": "string",           // Name of the file
            "size": "number",           // Size of the file in bytes
            "updated": "string",      // Date and time when the item was updated
            "path": "string",           // Location of the file
            "parent": {                 // Parent folder details (if available)
              "id": "string",
              "name": "string"
            }
          }
        ],
        "pagination": {
          "totalItems": 100,           // Total number of items
          "totalPages": 5,             // Total number of pages
          "currentPage": 2,            // Current page number
          "pageSize": 20               // Number of items per page
        }
      }
    ```
    - exemple of response:
    ```...```
   
  - 401 Unauthorized
    - Authentication Failure: User is not authenticated or token is invalid.
    ```json 
        {
            "status": "error",
            "error_code": "401",
            "message": "User not authenticated."
        }
    ```
  - 400 Bad Request
    - Invalid query parameters: The query parameters are missing or invalid.
    ```json 
        {
          "status": "error",
          "message": "Invalid query parameters"
        }
    ```
  - 500 Internal Error
    - Server Error: Internal server error occurs while processing the request.
    ``` json
        {
            "status": "error",
            "error_code": "500",
            "message": "An unexpected error occurred. Please try again later."
        }
    ```

docs/DisplayUploadedFilesInStorage.md
```diff
@@ -0,0 +1,171 @@
...
+
+
+### Frontend design
+
+- Is this a hook or a mutation? 
+  - This is a hook because we are fetching data from the backend to display in the frontend UI.
```
ReidyT, line 142:
```suggestion
- A hook is used in the frontend to fetch the data from the backend.
```

docs/DisplayUploadedFilesInStorage.md
```diff
@@ -0,0 +1,171 @@
...
+        }
+        ```
+
+
+### Frontend design
+
+- Is this a hook or a mutation? 
+  - This is a hook because we are fetching data from the backend to display in the frontend UI. 
+- What are the parameters of the hook?
+  -  Parameters for the hook will include:
```
ReidyT, line 144:
```suggestion
-  Parameters for the hook:
```

docs/DisplayUploadedFilesInStorage.md
```diff
@@ -0,0 +1,171 @@
...
+      - page: The page number to fetch.
+      - pageSize: The number of items per page.
+
+      These parameters will map to the query string parameters used in the backend endpoint (`/members/current/storage/list?page=2&pageSize=20`).
+
+### Backend design
+
+- Where should this go? Is this a new service? Does it extend an already existing service?
+  - This functionality can extend an existing service that manages items. If no such service exists, a new service specifically for managing items (files and folders) should be created.
```
ReidyT, line 153: I don't think this is useful, I'd rather precise in which service the new feature will be implemented if we already know that (for example in the `MemberService`, `ItemService` or any other service).

If we don't know for now, just let like that.

ReidyT (commented):
Thanks for this PR, good job 🤠! I've let a few suggestions for you that you are free to apply or not. I'm open to discussion and interested in what you (as well as the rest of the team) think. I haven't receive any information about this PR, so I don't know if you have any constraints for this document.

spaenleh (approved):

PR #292, opened 2024-07-19
fix: add storage intercept and test

spaenleh:
In this PR:
- add a small test to check the storage interface
- add type annotations to the `setUpApi` command
- remove unused intercepts

### DIFF
--- a/cypress/e2e/storage.cy.ts
+++ b/cypress/e2e/storage.cy.ts
@@ -7,13 +7,13 @@ import { CURRENT_MEMBER } from '../fixtures/members';
 
 describe('Storage', () => {
   it('Display storage interface', () => {
-    const storageAmount = 698789;
-    cy.setUpApi({ currentMember: CURRENT_MEMBER, storageAmount });
+    const storageAmountInBytes = 698789;
+    cy.setUpApi({ currentMember: CURRENT_MEMBER, storageAmountInBytes });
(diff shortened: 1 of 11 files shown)


cypress/e2e/storage.cy.ts
```diff
@@ -0,0 +1,19 @@
+import { formatFileSize } from '@graasp/sdk';
+
+import { STORAGE_PATH } from '@/config/paths';
+import { STORAGE_BAR_LABEL_ID } from '@/config/selectors';
+
+import { CURRENT_MEMBER } from '../fixtures/members';
+
+describe('Storage', () => {
+  it('Display storage interface', () => {
+    const storageAmount = 698789;
```
ReidyT, line 10:
```suggestion
    const storageAmountInBytes = 698789;
```

cypress/support/commands.ts
```diff
@@ -34,16 +49,15 @@ Cypress.Commands.add(
     postAvatarError = false,
     updatePasswordError = false,
     updateEmailError = false,
+    storageAmount = 10000,
```
ReidyT, line 52:
```suggestion
    storageAmountInBytes = 10000,
```

ReidyT (approved):
Thanks for this PR! LGTM 🚀. I'd just suggest to specify the unit for the amount directly in the variable or param (like `storageAmountInBytes`).

PR #299, opened 2024-07-22
feat: add email from query string in the validate email page

### LINKED ISSUE #286: Follow-up #275
Given the reviews on https://github.com/graasp/graasp-account/pull/275, the following changes need to be implemented:
### Give feedback to the user when they update the email

When entering a new email, the user needs to get feedback that the change was ok and the new email should be displayed in the interface.

### Give more information on the validation page

When users access the validation page, they should be able to see to which email their account will be updated. 
For this we can have multiple ways:
- add the email in plain text in the url query-string (can be error prone, as you could fake the email by changing the query string)
- have an endpoint that replies with the email associated with the request token (more secure)

spaenleh:
In this PR I use the email that is provided in the url when validating the new email to display more information to the user.
Screenshot 2024-07-22 at 10 45 55
fix #286


src/pages/EmailChangeValidationScreen.tsx
```diff
@@ -102,6 +104,7 @@ const Content = (): JSX.Element => {
         <Card>
           <Stack direction="column" alignItems="center" gap={1} p={2}>
             <Typography>{translate('VALIDATE_EMAIL_TEXT')}</Typography>
+            <Typography fontWeight="bold">{newEmail}</Typography>
             <Button
```
pyphilia, line 108: Should the button be `contained`? Just a suggestion.

pyphilia (approved):

src/pages/EmailChangeValidationScreen.tsx
ReidyT: I agree with Kim, (at least it's a good suggestion ^^)

ReidyT (approved):
Thanks for this PR, LGTM !

src/pages/EmailChangeValidationScreen.tsx
spaenleh: 👍

PR #330, opened 2024-08-20
fix: remove page size option that exceeds the maximum allowed in storage table

mariembencheikh:
Since the max page size is set to 50 in the backend so I deleted the select of 100 items


spaenleh (approved):
Looks good ! Thank you for the quick fix.

PR #333, opened 2024-08-22
feat: add 404 page

mariembencheikh:
close to #281 

image

### DIFF
--- a/cypress/e2e/notFoundPage.cy.ts
+++ b/cypress/e2e/notFoundPage.cy.ts
@@ -1,23 +1,28 @@
+import {
+  GO_TO_HOME_LINK_ID,
+  NOT_FOUND_MESSAGE_ID,
+  NOT_FOUND_TEXT_ID,
+} from '@/config/selectors';
+
 import { BOB } from '../fixtures/members';
 
 describe('404 Page Test', () => {
   beforeEach(() => {
     cy.setUpApi({ currentMember: BOB });
+    cy.visit('/non-existing-page');
   });
   it('should display 404 page for non-existing routes', () => {
-    cy.visit('/non-existing-page');
+    cy.get(`#${NOT_FOUND_TEXT_ID}`).should('contain', 'Page not found');
 
-    cy.url().should('include', '/non-existing-page');
-
-    cy.get('h1').contains('404 Page not found').should('be.visible');
-
-    cy.contains('Sorry, we could not find this page.').should('be.visible');
-    cy.get('a').contains('Go to Home').should('have.attr', 'href', '/');
+    cy.get(`#${NOT_FOUND_MESSAGE_ID}`).should(
+      'contain',
+      'Sorry, we could not find this page.',
+    );
   });
 
   it('should navigate to the home page when clicking "Go to Home"', () => {
     cy.visit('/non-existing-page');
-    cy.get('a').contains('Go to Home').click();
+    cy.get(`#${GO_TO_HOME_LINK_ID}`).click();
 
     // Check if the URL is the home page
     cy.url().should('eq', `${Cypress.config().baseUrl}/`);
--- a/cypress/e2e/notFoundPage.cy.ts
+++ b/cypress/e2e/notFoundPage.cy.ts
@@ -1,0 +1,25 @@
+import { BOB } from '../fixtures/members';
+
+describe('404 Page Test', () => {
+  beforeEach(() => {
+    cy.setUpApi({ currentMember: BOB });
+  });
+  it('should display 404 page for non-existing routes', () => {
+    cy.visit('/non-existing-page');
+
+    cy.url().should('include', '/non-existing-page');
+
+    cy.get('h1').contains('404 Page not found').should('be.visible');
+
+    cy.contains('Sorry, we could not find this page.').should('be.visible');
+    cy.get('a').contains('Go to Home').should('have.attr', 'href', '/');
+  });
+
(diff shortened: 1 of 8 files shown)


src/pages/NotFoundPage.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const NotFoundPage = (): JSX.Element => {
+  const { t } = useAccountTranslation();
+
+  return (
+    <div style={{ textAlign: 'center', padding: '50px' }}>
+      <h1>{t('NOT_FOUND_PAGE_TEXT')}</h1>
+      <p>{t('NOT_FOUND_Page_MESSAGE')}</p>
+      <Link to="/">{t('GO_TO_HOME_TEXT')}</Link>
+    </div>
```
spaenleh, line 14: Why not using the Mui components ?

src/langs/en.json
```diff
@@ -111,5 +111,8 @@
   "EXPORT_INFORMATIONS_DESCRIPTION": "You can download a copy of your data.",
   "EXPORT_INFORMATIONS_BUTTON_TEXT": "Download your data",
   "EXPORT_SUCCESS_MESSAGE": "You have successfully requested your data. You will receive an e-mail with your data in a few minutes.",
-  "EXPORT_ERROR_MESSAGE": "An unexpected error occurred"
+  "EXPORT_ERROR_MESSAGE": "An unexpected error occurred",
+  "NOT_FOUND_PAGE_TEXT": "404 Page not found",
```
spaenleh, line 115: Actually I don't know if we should show the 404 error code, as this is not really something that users should see no ?
```suggestion
  "NOT_FOUND_PAGE_TEXT": "Page not found",
```

src/langs/en.json
```diff
@@ -111,5 +111,8 @@
...
+  "NOT_FOUND_Page_MESSAGE": "Sorry, we could not find this page.",
```
spaenleh, line 116:
```suggestion
  "NOT_FOUND_PAGE_MESSAGE": "Sorry, we could not find this page.",
```

src/pages/NotFoundPage.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import { useAccountTranslation } from '@/config/i18n';
+
+const NotFoundPage = (): JSX.Element => {
+  const { t } = useAccountTranslation();
+
+  return (
+    <div style={{ textAlign: 'center', padding: '50px' }}>
+      <h1>{t('NOT_FOUND_PAGE_TEXT')}</h1>
+      <p>{t('NOT_FOUND_Page_MESSAGE')}</p>
```
spaenleh, line 12: There is a case issue in your translation string.

cypress/e2e/notFoundPage.cy.ts
```diff
@@ -0,0 +1,25 @@
+import { BOB } from '../fixtures/members';
+
+describe('404 Page Test', () => {
+  beforeEach(() => {
+    cy.setUpApi({ currentMember: BOB });
+  });
+  it('should display 404 page for non-existing routes', () => {
+    cy.visit('/non-existing-page');
```
spaenleh, line 8: Put this in the beforeEach

cypress/e2e/notFoundPage.cy.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    cy.get('h1').contains('404 Page not found').should('be.visible');
+
+    cy.contains('Sorry, we could not find this page.').should('be.visible');
+    cy.get('a').contains('Go to Home').should('have.attr', 'href', '/');
+  });
+
+  it('should navigate to the home page when clicking "Go to Home"', () => {
+    cy.visit('/non-existing-page');
+    cy.get('a').contains('Go to Home').click();
```
spaenleh, line 20: Here, are you using the `contains` to assert that the text is correct ? Or is it used as a selecctor to narrow the selected element ?

cypress/e2e/notFoundPage.cy.ts
```diff
@@ -0,0 +1,25 @@
+import { BOB } from '../fixtures/members';
+
+describe('404 Page Test', () => {
+  beforeEach(() => {
+    cy.setUpApi({ currentMember: BOB });
+  });
+  it('should display 404 page for non-existing routes', () => {
+    cy.visit('/non-existing-page');
+
+    cy.url().should('include', '/non-existing-page');
```
spaenleh, line 10: I am not sure you need to ensure that the url is correct ... unless this is supposed to check that there is no automatic redirect. If this is the case, please state the purpose of this check. As is I find it is not necessary and would probably remove it if there is not a reason stated for why it is there.

cypress/e2e/notFoundPage.cy.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    cy.get('h1').contains('404 Page not found').should('be.visible');
```
spaenleh, line 12: Same question here about the use of `.contains`. Do you use it as an assertion or a selector ?

spaenleh (changes requested):
Thank you for the work.

I have a general question about the tests, you use [.contains](https://docs.cypress.io/api/commands/contains#__docusaurus_skipToContent_fallback) and I am not sure if you did not mean to use `.should('contain', '...')` for asserting instead.

I also see that the style of the assertions and how to select the elements is not really in line with how we used to do it in other tests. You do not use ids and proper selectors. Is this intentional ? I would prefer to have a consistent style across tests and source code.

spaenleh (approved):

PR #355, opened 2024-09-20
fix: update dependencies and fix issues

spaenleh:


src/components/main/MemberPreferences.tsx
```diff
@@ -17,14 +18,20 @@ import BorderedSection from '../layout/BorderedSection';
...
 import MemberProfileItem from './MemberProfileItem';
 
-const MemberPreferences = (): JSX.Element => {
+const MemberPreferences = (): JSX.Element | false => {
   const { data: member } = hooks.useCurrentMember();
+
+  const { t } = useAccountTranslation();
+  const [isEditing, setIsEditing] = useState(false);
+
+  if (!member || member?.type === AccountType.Guest) {
```
pyphilia, line 27: I would rather check the member is an `Individual` than not a `Guest`.

src/components/main/MemberPreferences.tsx
```diff
@@ -17,14 +18,20 @@ import BorderedSection from '../layout/BorderedSection';
...
+    return false;
```
pyphilia, line 28: Should we return `null`?

src/modules/profile/PersonalInformation.tsx
```diff
@@ -24,6 +26,10 @@ const PersonalInformation = (): JSX.Element | false => {
 
   const onClose = () => setIsEditing(false);
 
+  if (member?.type === AccountType.Guest) {
+    return false;
+  }
```
pyphilia, line 31: Same as above.

pyphilia (changes requested):
Thanks! 🩵

src/components/main/MemberPreferences.tsx
spaenleh: I started returning `false` instead of `null` after reading [this article](https://medium.com/@davidkelley87/stop-using-return-null-in-react-a2ebf08fc9cd)

ReidyT: I'm a bit mitigate with this article (check the comments, there is a lot of people that disagree with this practice). Even in the react's documentation they explain how to do conditional rendering using `null` (https://react.dev/learn/conditional-rendering#conditionally-returning-nothing-with-null).

ReidyT (changes requested):
Thanks for this PR, LGTM! Maybe we should just discuss about returning `null` or `false`? I answered directly in the comment 😄

src/components/main/MemberPreferences.tsx
spaenleh: I looked a bit more and it is true that using `false` can actually have unintended consequences. Thank you for pointing it out. I will revert to using `null` instead. 👍

src/components/main/MemberPreferences.tsx
```diff
@@ -17,14 +18,21 @@ import BorderedSection from '../layout/BorderedSection';
...
+const MemberPreferences = (): JSX.Element | null => {
   const { data: member } = hooks.useCurrentMember();
+
+  const { t } = useAccountTranslation();
+  const [isEditing, setIsEditing] = useState(false);
+
+  // in case there is no member or they are not of the individual type, we render nothing
+  if (!member || member?.type !== AccountType.Individual) {
+    return null;
+  }
```
pyphilia, line 30: If this is a page, should we redirect? 
If this is a component, should we enforce that the member exists with the props? Or maybe show an error here?

pyphilia (approved):

ReidyT (approved):
LGTM 🐦‍🔥

src/components/main/MemberPreferences.tsx
spaenleh: This is a component but should I pass it the member as props ? I do not really like the idea of having to prop drill it. 

Maybe since we have the need for it in a few places, we could have the api return the member only if we request a "real member" ? maybe it should be a different endpoint ? So it would return noting if the user is a guest. And we could remove the checks in the frontend ? @ReidyT @pyphilia @Defferrard

PR #356, opened 2024-09-20
fix: do not allow special characters in username

MatveyK:
Relates to https://github.com/graasp/graasp/pull/1406

### DIFF
--- a/src/langs/it.json
+++ b/src/langs/it.json
@@ -76,6 +76,7 @@
   "DESTRUCTIVE_SETTINGS_DETAILS": "Tieni presente che le azioni qui descritte hanno conseguenze potenzialmente irreversibili. Assicurati di esaminare attentamente eventuali restrizioni prima di procedere, e di agire solo quando hai certezza.",
   "USERNAME_EMPTY_ERROR": "Il campo non può essere vuoto",
   "USERNAME_LENGTH_ERROR": "Il nome utente deve contenere tra {{min}} e {{max}} caratteri",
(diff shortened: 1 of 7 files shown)


src/langs/it.json
```diff
@@ -115,4 +116,4 @@
   "NOT_FOUND_PAGE_TEXT": "Pagina non trovata",
   "NOT_FOUND_PAGE_MESSAGE": "Spiacenti, non siamo riusciti a trovare questa pagina.",
   "GO_TO_HOME_TEXT": "Vai a Home"
-}
+}
```
spaenleh, line 119: Can you check that your editor is setup to add an empty new line at the end of files ?
```
  "files.trimTrailingWhitespace": true
```

spaenleh (changes requested):
Thank you for this fix, can you add a test that verifies that when inputing a forbidden character the user sees the error message ?

spaenleh:
@MatveyK Also see about the issue raised by sonarCloud. Maybe there is way to make the regex simpler ? Not sure we need backtracking in there ?

src/langs/it.json
MatveyK: Added.

MatveyK:
Regex simplified and test added.

pyphilia (approved):
LGTM, same comment as in https://github.com/graasp/graasp-auth/pull/457#pullrequestreview-2328448199

spaenleh (approved):

PR #364, opened 2024-10-08
feat: use password status, react hook form

### LINKED ISSUE #359: Use new password queries (post, patch and status)
There is a new endpoint to get the password status of the current member which allows to know if they have set a password already or not. 

New endpoints also exist to:
1. create a password when the member does not have one already
2. update the password when they want to change it

We need to use these new endpoints to provide a better user experience for users that will want to add or update their passwords. Currently the experience is really not great because it only uses the patch endpoint and thus users need to provide a password in the current password field even though they do not have one set...

pyphilia:
Show correct informations based on whether the user has a password set. 
Use POST instead of PATCH when the user does not have a password.
Simplify password edition with `react-hook-form`.

https://github.com/user-attachments/assets/601d1606-a748-4dab-b69a-07cf0209718d

https://github.com/user-attachments/assets/e71864f8-35bf-49de-b2e9-755cb143252b

close #359


cypress/support/server.ts
```diff
@@ -253,6 +253,22 @@ export const mockUpdatePassword = (shouldThrowError: boolean): void => {
...
+    {
+      method: HttpMethod.Post,
+      pathname: `/password`,
+    },
+    ({ reply }) => {
+      if (shouldThrowError) {
+        return reply({ statusCode: StatusCodes.BAD_REQUEST });
+      }
+
+      return reply('create password');
```
spaenleh, line 267: This should match the reply from the backend: 204.

package.json
```diff
@@ -14,10 +14,10 @@
     "@emotion/cache": "11.13.1",
     "@emotion/react": "11.13.3",
     "@emotion/styled": "11.13.0",
-    "@graasp/query-client": "4.0.1",
+    "@graasp/query-client": "github:graasp/graasp-query-client#password-status-hook",
```
spaenleh, line 17: Reminder

src/config/notifier.ts
```diff
@@ -93,10 +88,10 @@ export default ({
 
   // error notification
   if (payload?.error && message) {
-    toast.error(i18n.t(message, { ns: ACCOUNT_NAMESPACE }));
+    toast.error(i18n.t(message, { ns: namespaces.messages }));
```
spaenleh, line 91: Is this change correct ? this now translates the message using the translations namespace.

src/langs/en.json
```diff
@@ -40,6 +40,7 @@
   "PROFILE_AVATAR_CURRENT_ALT": "Profile Picture",
   "PASSWORD_TITLE": "Password",
   "PASSWORD_SETTINGS_INFORMATION": "Update your password by clicking on the 'Edit' button",
+  "NEW_PASSWORD_SETTINGS_INFORMATION": "Add a password by clicking on the 'Configure' button",
```
spaenleh, line 43: This string does not flow the scope convention: 
```suggestion
  "PASSWORD_SETTINGS_INFORMATION_NEW_PASSWORD": "Add a password by clicking on the 'Configure' button",
```

src/modules/profile/password/EditPassword.tsx
```diff
@@ -1,168 +1,238 @@
...
+        updatePasswordError.response?.data.name ??
+          FAILURE_MESSAGES.UNEXPECTED_ERROR,
+      )
+    : null;
+  const createNetworkError = axios.isAxiosError(createPasswordError)
+    ? translateMessage(
+        createPasswordError.response?.data.name ??
+          FAILURE_MESSAGES.UNEXPECTED_ERROR,
+      )
+    : null;
```
spaenleh, line 111: I think it is interesting, but do we need to be so specific about the issue ? See what I did in auth, When an error occurred I tried to point the user to a valid action to perform and usually deduce that a failure could be due to a single cause.

src/modules/profile/password/EditPassword.tsx
```diff
@@ -1,168 +1,238 @@
...
+          <Stack spacing={2}>
+            {passwordStatus?.hasPassword && (
+              <Box>
+                <TextField
+                  required
+                  id={PASSWORD_INPUT_CURRENT_PASSWORD_ID}
+                  label={t('PASSWORD_SETTINGS_CURRENT_LABEL')}
+                  variant="outlined"
+                  size="small"
+                  type="password"
```
spaenleh, line 134: Allow to see the value of the field in clear text ?

src/modules/profile/password/EditPassword.tsx
```diff
@@ -1,168 +1,238 @@
...
+                  error={Boolean(currentPasswordErrorMessage)}
+                  helperText={
+                    currentPasswordErrorMessage &&
+                    t(currentPasswordErrorMessage)
+                  }
+                  autoComplete="current-password"
```
spaenleh, line 140: What is this ?

src/modules/profile/password/EditPassword.tsx
```diff
@@ -1,168 +1,238 @@
...
+                  // eslint-disable-next-line react/jsx-props-no-spreading
```
spaenleh, line 141: Should we disable this rule as we benefit from it ?

src/modules/profile/password/EditPassword.tsx
```diff
@@ -1,168 +1,238 @@
...
-            variant="outlined"
+    <Box component="form" onSubmit={handleSubmit(onSubmit)}>
+      <BorderedSection
+        id={PASSWORD_EDIT_CONTAINER_ID}
+        title={t('PASSWORD_SETTINGS_TITLE')}
+      >
+        <Typography variant="body1">
+          {t('PASSWORD_SETTINGS_CONFIRM_INFORMATION')}
+        </Typography>
+        {passwordStatus ? (
```
spaenleh, line 124: Should this component be split to ease the flow. Like have a wrapper that fetches the password status and the the rendering of the content in another component ? I find it cleared to split when the components rendered conditionally are big. For a "one line" it is ok to have them inline, but with bigger trees I find having explicit control flow (with `if`) better for readability.

spaenleh (changes requested):
Thank you for this feature ! Code look good with the usage of the `react-hook-form` library ! 

I have left a few comments about readability and consistency, let me know what you think !

ReidyT (commented):
Thanks for this PR! I haven't check the code for now, I just have questions/comments about the feature's behaviour.

1. When setting a new password, is it possible to display the error live (while entering the new password) instead of when saving?
2. Same comment for when confirming the password (displaying in live that the two passwords are different).
3. Should we check the current password for weaknesses?

src/langs/constants.ts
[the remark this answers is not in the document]
pyphilia: I did it mainly because I used it in tests. I prefer having constants in this case.

src/modules/profile/password/EditPassword.tsx
pyphilia: It's for auto complete from the user agent, but we can deal with that later, I'll revert.

src/modules/profile/password/EditPassword.tsx
pyphilia: I don't think it's a good practice, I would rather have the eye icon.

src/modules/profile/password/EditPassword.tsx
spaenleh: Okay that is fine, I was just curious

ReidyT (commented):
Also I agree with what Basile said, otherwise it looks good to me.

src/modules/profile/password/EditPassword.tsx
spaenleh: Okay, the eye is fine for me 👍

pyphilia:
Thanks for the reviews!

The fields are checked live if you submitted once. I think it's a nice pattern, otherwise you start by checking that an empty password is not strong.. current password is checked as strong password as well.

src/modules/profile/password/EditPassword.tsx
pyphilia: This is mainly to deal with backend errors. Especially "current password is not correct" error. Otherwise it might show "unexpected error instead". So I prefer to be exhaustive. 
I'll refactor the component, so it won't have both create and update at the same time, might be easier to read.

src/config/notifier.ts
pyphilia: Might be wrong, but the notifier is only used by query client, which only uses messages. So it was wrong from the very beginning I think?

src/modules/profile/password/CreatePassword.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  const createNetworkError = axios.isAxiosError(createPasswordError)
+    ? translateMessage(
+        createPasswordError.response?.data.name ??
+          FAILURE_MESSAGES.UNEXPECTED_ERROR,
+      )
+    : null;
+
+  return (
+    <Box component="form" onSubmit={handleSubmit(onSubmit)}>
```
spaenleh, line 89: I think this form should be inside the `BorderedSection` component

spaenleh: Actually the "form" should be set on the Stack that is below the title. remove the box component.

src/modules/profile/password/DisplayPassword.tsx
```diff
@@ -1,20 +1,25 @@
...
 } from '@/config/selectors';
+import { ACCOUNT } from '@/langs/constants';
 
+import CreatePassword from './CreatePassword';
 import EditPassword from './EditPassword';
 
 const DisplayPassword = (): JSX.Element => {
   const { t } = useAccountTranslation();
   const [isEditing, setIsEditing] = useState(false);
+  const { data: passwordStatus, isLoading: isPasswordStatusLoading } =
```
spaenleh, line 21: Depending on if you want only the initial load or if you want all the loadings you should use `isPending`. `isLoading` has changed semantics in v5.

src/modules/profile/password/EditPassword.tsx
```diff
@@ -1,168 +1,192 @@
...
+  const updateNetworkError = axios.isAxiosError(updatePasswordError)
+    ? translateMessage(
+        updatePasswordError.response?.data.name ??
+          FAILURE_MESSAGES.UNEXPECTED_ERROR,
+      )
+    : null;
 
-  const hasChanged =
-    newPasswordError === null && newPassword === confirmPassword;
+  const networkError = updateNetworkError;
```
spaenleh, line 100: Why do you rename this ? Do you need it ?

src/modules/profile/password/EditPassword.tsx
```diff
@@ -1,168 +1,192 @@
...
-            helperText={newPasswordError}
-            onChange={handleNewPasswordInput}
-            type="password"
-            id={PASSWORD_INPUT_NEW_PASSWORD_ID}
-          />
-          <TextField
-            required
-            label={t('PASSWORD_SETTINGS_NEW_CONFIRM_LABEL')}
-            variant="outlined"
+    <Box component="form" onSubmit={handleSubmit(onSubmit)}>
```
spaenleh, line 103: Same comment here, set the form on the Stack on line 111.

spaenleh (approved):
Looks good, I would just like to move the foms down a bit. You may have to include the buttons in the stacks, but I think it will be good anyways.

ReidyT (approved):
LGTM

PR #376, opened 2024-10-16
fix: update ts config

spaenleh:
In this PR: 
- update ts config
- remove paths in tests as they do not work well with Cypress see #378


ReidyT (approved):
LGTM! Maybe we can create another issue to try to fix the path alias?

spaenleh:
Ref #381 for the path alias changes

PR #377, opened 2024-10-16
fix: update eslint config

spaenleh:
In this PR: 
- update eslint config and use v9
- remove unused util files


pyphilia (approved):
LGTM 🍸

PR #408, opened 2024-10-30
fix: updated lang is not shown in preferences when switching lang

spaenleh:
In this PR: show the right language in the switch when editing the member language.

Previously it was staying on the current language.
It now changes to show the selected language when the select closes.


pyphilia (approved):
LGTM!

PR #411, opened 2024-10-30
fix: use the script from cloud and send to our own

spaenleh:
The issue is that the request to get the script does not allow Cross origin requests, so the browser blocks it. 

We can load the script from the cloud.umami.is website and send data to our own instance.

It should work.

It works.


ReidyT (approved):
Nice fix!

PR #448, opened 2024-11-15
fix: recaptcha script was not loaded in the html

spaenleh:


ReidyT (approved):
Looks really good! Nice job

PR #460, opened 2024-11-19
fix: move recaptcha to auth layout so we do not track on all pages

### LINKED ISSUE #459: Recaptcha script is loaded on all pages
Ther ecaptcha script is added to the `index.html` by the build process handeld by vite. This enables the script on **all** pages of the app. This is not what we want. We should only load the script on page for `/auth/...`.

spaenleh:
In this PR I update the way we load the recaptcha script.

fixes #459 

## Previously

Recaptcha script was added to the main `index.html`  file at build with the help of a vite plugin that used the transformHtml feature.
It allowed to stop the build if the variable was not present but it was marked as required, making common "missing env var" errors a non-issue.

## The bug

With the previous implementation, the script was loaded when you opened **any** page of the app, since it was in the main `index.html` that was used by the SPA.
This means that the recaptcha badge was shown on **all** pages. Which it not desirable because:
- it can block some elements
- it tracks users un-necessarily

## What was changed

In order to **only** load the recaptcha script on the auth pages, we rely on `react-helmet-async` to dynamically add the script tag to the pages that are under the `/auth` path. 
It also correctly removes the tag when we visit other pages not on this path. 
This gives the expected behaviour of loading the script only for the pages that require it.


ReidyT (approved):
LGTM! So now we don't detect the missing recaptcha env var anymore right?

spaenleh:
@ReidyT Yes this is the only downside of the new implementation. Since the env var is not references at build time, there is nothing stopping us from building without it. I can add a check in the build that fails if it is missing and it should be prensent (building in prod for example), but not sure we need it ?

pyphilia (approved):

ReidyT:
> @ReidyT Yes this is the only downside of the new implementation. Since the env var is not references at build time, there is nothing stopping us from building without it. I can add a check in the build that fails if it is missing and it should be prensent (building in prod for example), but not sure we need it ?

I think it was an interesting thing to be able to detect at build time that the env is missing, but it's not a big problem to lose that functionality.

PR #468, opened 2024-11-21
fix: update adornemnt and use slotprops

spaenleh:
In this PR:
- fix email adornment in register from mail to user icon
- fix `InputProps` and use slotProps instead


src/modules/auth/components/common/Adornments.tsx
```diff
@@ -1,12 +1,19 @@
 import { InputAdornment } from '@mui/material';
```
pyphilia, line 1: Rename the file to `adornments.tsx`?

pyphilia (approved):

PR #474, opened 2024-11-21
feat: add react-hook-form on register

spaenleh:
In this PR:
- I add react-hook form to the register form
- fix some small issues with url passed from the landing page

### DIFF
--- a/src/modules/auth/components/register/Register.tsx
+++ b/src/modules/auth/components/register/Register.tsx
@@ -41,8 +41,6 @@ import { FormHeader } from '../common/FormHeader';
 import { StyledTextField } from '../common/StyledTextField';
 import { EmailAdornment, NameAdornment } from '../common/adornments';
 
-const { SIGN_IN_LINK_TEXT, NAME_FIELD_LABEL, SIGN_UP_BUTTON } = AUTH;
-
 type RegisterInputs = {
   name: string;
   email: string;
@@ -50,11 +48,11 @@ type RegisterInputs = {
   userHasAcceptedAllTerms: boolean;
 };
 
-function EnableAnalyticsForm({
-  control,
-}: {
+type FormElementProps = {
   control: Control<RegisterInputs>;
-}): JSX.Element {
+};
+
+function EnableAnalyticsForm({ control }: FormElementProps): JSX.Element {
   const { field } = useController({ control, name: 'enableSaveActions' });
   const { t } = useTranslation(NS.Auth);
 
@@ -79,11 +77,7 @@ function EnableAnalyticsForm({
   );
 }
 
-export function AgreementForm({
-  control,
-}: {
-  control: Control<RegisterInputs>;
-}) {
+export function AgreementForm({ control }: FormElementProps) {
   const { t } = useTranslation(NS.Auth);
   const {
     field,
@@ -91,9 +85,9 @@ export function AgreementForm({
   } = useController({
     control,
     name: 'userHasAcceptedAllTerms',
-    rules: { required: 'yes' },
+    rules: { required: t('USER_AGREEMENTS_REQUIRED') },
   });
-  const hasError = Boolean(errors.userHasAcceptedAllTerms?.message);
+  const validationError = errors.userHasAcceptedAllTerms?.message;
 
   return (
     <>
@@ -102,7 +96,7 @@ export function AgreementForm({
           <Checkbox
             checked={field.value}
             onChange={(_, checked) => field.onChange(checked)}
-            color={hasError ? 'error' : 'primary'}
+            color={validationError ? 'error' : 'primary'}
             data-cy={REGISTER_AGREEMENTS_CHECKBOX_ID}
             size="small"
           />
@@ -111,7 +105,7 @@ export function AgreementForm({
           <Typography
             display="inline"
             fontSize="small"
-            color={hasError ? 'error' : 'default'}
+            color={validationError ? 'error' : 'default'}
           >
             <Trans
               t={t}
@@ -124,9 +118,9 @@ export function AgreementForm({
           </Typography>
         }
       />
-      {hasError && (
+      {validationError && (
         <Typography variant="caption" color="error">
-          {t('USER_AGREEMENTS_REQUIRED')}
+          {validationError}
         </Typography>
       )}
     </>
@@ -138,7 +132,7 @@ type RegisterProps = {
     url?: string;
     invitationId?: string;
   };
-  initialData: {
+  initialData?: {
     name?: string;
     email?: string;
   };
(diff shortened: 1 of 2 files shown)


public/locales/en/common.json
```diff
@@ -38,5 +38,12 @@
   },
   "ERRORS": {
     "UNEXPECTED": "An unexpected Error occurred"
+  },
+  "FIELD_ERROR": {
+    "REQUIRED": "This field is required",
+    "USERNAME_MIN_LENGTH": "User name must be longer than {{min}} characters",
+    "USERNAME_MAX_LENGTH": "User name must be less than {{max}} characters",
```
pyphilia, line 45: I've seen the same messages in auth, duplicate?

src/routes/auth/register.tsx
```diff
@@ -1,27 +1,74 @@
...
+    </LeftContentContainer>
+  ),
 });
 
-function RegisterPage() {
+function RegisterWithInvitation() {
   const search = Route.useSearch();
+  const { t } = useTranslation(NS.Auth);
+
+  const { data: invitation, isPending: isLoadingInvitations } =
```
pyphilia, line 35:
```suggestion
  const { data: invitation, isPending: isLoadingInvitation } =
```

src/modules/auth/components/register/Register.tsx
```diff
@@ -1,74 +1,174 @@
...
+  enableSaveActions: boolean;
+  userHasAcceptedAllTerms: boolean;
+};
+
+function EnableAnalyticsForm({
+  control,
+}: {
+  control: Control<RegisterInputs>;
+}): JSX.Element {
+  const { field } = useController({ control, name: 'enableSaveActions' });
```
pyphilia, line 58: This is a very big file, within `Register.tsx`. Is it really worth? If you use `useFormContext` with subtypes maybe it would be possible to split?

src/modules/auth/components/register/Register.tsx
```diff
@@ -81,145 +181,126 @@ export function Register({ search }: RegisterProps) {
...
-    }
+    // navigate to success path
+    navigate({
+      to: '/auth/success',
+      search: { email, back: location.pathname },
+    });
   };
 
+  const nameError = errors.name?.message satisfies string | undefined;
+  const emailError = errors.email?.message satisfies string | undefined;
```
pyphilia, line 220: Necessary typing?

src/modules/auth/components/register/Register.tsx
```diff
@@ -81,145 +181,126 @@ export function Register({ search }: RegisterProps) {
...
+                min: MIN_USERNAME_LENGTH,
+              }),
+            },
+            validate: (name) =>
+              MemberConstants.USERNAME_FORMAT_REGEX.test(name.trim()) ||
+              translateCommon('USERNAME_SPECIAL_CHARACTERS'),
+          })}
+          placeholder={t(NAME_FIELD_LABEL)}
+          // todo: Should we not allow users to change their name when creating a count from an invitation ?
+          // disabled={Boolean(invitation?.name)}
```
pyphilia, line 263: Remove?

src/modules/auth/components/register/Register.tsx
```diff
@@ -81,145 +181,126 @@ export function Register({ search }: RegisterProps) {
...
+        />
+        <StyledTextField
+          id={EMAIL_SIGN_UP_FIELD_ID}
+          slotProps={{
+            input: {
+              startAdornment: EmailAdornment,
+            },
+          }}
+          variant="outlined"
+          // type="email"
```
pyphilia, line 273: keep?

src/modules/auth/components/register/Register.tsx
```diff
@@ -81,145 +181,126 @@ export function Register({ search }: RegisterProps) {
...
-        {t(SIGN_IN_LINK_TEXT_BUTTON)}
-      </Link>
+        <ErrorDisplay error={registerError} />
+        <LoadingButton
+          type="submit"
+          variant="contained"
+          id={REGISTER_BUTTON_ID}
+          loading={isLoadingSignUp || isLoadingMobileSignUp}
+          fullWidth
+          disabled={Boolean(Object.keys(errors).length > 0)}
```
pyphilia, line 296: What I did in builder is the following, isSubmitted being necessary because the initial value before submitting is false. In your case it's the same since `errors` is computed only after submitting. 
```suggestion
          disabled={isSubmitted && !isValid}
```

src/routes/auth/register.tsx
```diff
@@ -1,27 +1,74 @@
...
+
   return (
-    <LeftContentContainer>
-      <Register search={search} />
-    </LeftContentContainer>
+    <Stack gap={2} alignItems="center">
+      <Alert severity="warning" sx={{ maxWidth: '35ch' }}>
+        {t('INVITATION_NOT_FOUND_MESSAGE')}
+      </Alert>
+      <RegisterForm search={search} initialData={{}} />
```
pyphilia, line 62: Should initial data have default values to avoid using an empty object here?

src/routes/auth/register.tsx
```diff
@@ -1,27 +1,74 @@
...
   );
 }
+
+function RegisterPage() {
+  const search = Route.useSearch();
+  if (search.invitationId) {
+    return <RegisterWithInvitation />;
+  }
+
+  return <RegisterForm search={search} initialData={{}} />;
```
pyphilia, line 73: Same as above.

pyphilia (approved):
LGTM! It looks very good with react hooks form! 🚋

public/locales/en/common.json
spaenleh: Yes, this is why I added them to common. There is an open issue (https://github.com/graasp/client/issues/475) to refactor the account side to use the one in common.

src/modules/auth/components/register/Register.tsx
spaenleh: No, I will remove it.

src/modules/auth/components/register/Register.tsx
spaenleh: If we keep it, the form will use native validation (with the pop-up elements). When using react-hook form there are some HTML attributes we can not use (like `required` etc)

src/modules/auth/components/register/Register.tsx
spaenleh: This is interesting, but I tired it and since the button is disabled, it does not show the error message. So not quite what I was going for. But I learned something !

src/routes/auth/register.tsx
spaenleh: Yes that is a good idea !

PR #498, opened 2024-11-27
fix: issues highlighted

spaenleh:


public/locales/en/landing.json
```diff
@@ -275,5 +275,8 @@
       "POLICY": "Privacy Policy",
       "DISCLAIMER": "Disclaimer"
     }
+  },
+  "NAV": {
+    "GO_TO_GRAASP": "To the App"
```
pyphilia, line 280: Is it really clear?

src/components/ConstructionTruck.tsx
```diff
@@ -5,15 +5,16 @@ import { Box, styled } from '@mui/material';
 const MovingBox = styled(Box)({
   // animation: 'shake 500ms linear, move 40s linear infinite',
   animation: 'move 40s linear infinite',
-  position: 'fixed',
+  // position: 'relative',
   bottom: '0px',
-  right: '-100px',
+  right: '-200px',
+  // width: 'fit-content',
```
pyphilia, line 11: Remove comments.

src/modules/account/home/memberPicture/AvatarUploader.tsx
```diff
@@ -128,7 +127,7 @@ export function AvatarUploader(): JSX.Element {
           borderRadius={2}
           zIndex={theme.zIndex.drawer - 2}
         >
-          <ImageUpIcon color={theme.palette.secondary.light} />
+          <ImageUpIcon color="white" />
```
pyphilia, line 130: But shouldn't `white` be a color in the theme? It depends on the background color, which is also in the theme most of the time. Maybe text secondary?

src/modules/landing/footer/links.tsx
```diff
@@ -81,5 +81,7 @@ const MUILinkComponent = React.forwardRef<HTMLAnchorElement, MUILinkProps>(
 const CreatedLinkComponent = createLink(MUILinkComponent);
 
 export const InternalLink: LinkComponent<typeof MUILinkComponent> = (props) => {
-  return <CreatedLinkComponent preload={'intent'} {...props} />;
+  return (
+    <CreatedLinkComponent preload={'intent'} resetScroll={true} {...props} />
```
pyphilia, line 85:
```suggestion
    <CreatedLinkComponent preload="intent" resetScroll {...props} />
```

pyphilia (approved):

public/locales/en/landing.json
spaenleh: Not sure I am open to propositions.
Screenshot 2024-11-28 at 11 32 54

src/modules/account/home/memberPicture/AvatarUploader.tsx
spaenleh: In this instance it really depends on the color that is used for the background of the uploader, and in this instance we use the main color. We could use contrastText, but not sure it has the right meaning.

public/locales/en/landing.json
MatveyK: Maybe more like "Go to App ->", the action seems a bit clearer this way. And I would fill the color with primary colour to make it stand out, since it is the CTA button of this page.

PR #528, opened 2024-12-04
feat(ui): add Storybook

spaenleh:
In this PR:
- setup Storybook
- add stories to fix some common issues related to the setup of the app (add router decorator, useAuth mock etc)
- setup the ci to run component tests
- add stories to some of the basic components

### Changes and adaptations to the Translations config

> [!NOTE]
> **TLDR**: Translation files have moved from `/public/locales` to `src/locales` so we can import them in tests and they are copied at build time to the output directory. 
**TODO:** We will need to change (once again) the git-localize settings.

In order to make the play tests in storybook work with vitest, I had to use a different config for i18n in the tests.

In production, we use the fetch backend to load the translations from the web-server with an HTTP call.

This was not possible inside the tests since there was nothing to serve these files at that time. Storybook has an options to serve static assets but the play tests when executed in vitest do not start this static server.

Instead when running storybook and the vitest tests we load the resources and provide them with the `I18nextProvider` context to the cpomponent that will be rendered in the Story. This works.

In order to fix a warning about importing assets from the `public` dir with Javascript I moved the translation files back inside `src/locales` and added a vite plugin to copy them to the output directory on build.


.storybook/i18nTestInstance.ts
```diff
@@ -0,0 +1,29 @@
...
+import landing from '../src/locales/en/landing.json';
+import messages from '../src/locales/en/messages.json';
+import player from '../src/locales/en/player.json';
+
+i18n.use(initReactI18next).init({
+  lng: 'en',
+  fallbackLng: 'en',
+
+  // have a common namespace used around the full app
+  ns: ['auth', 'account', 'landing', 'messages', 'player'],
```
pyphilia, line 17: How will it behave if we forget to add a namespace here? Tho I don't think it will change often.

.storybook/preview.tsx
```diff
@@ -0,0 +1,116 @@
...
+      ],
+      // Property that specifies if the name of the item will be displayed
+      showName: true,
+      // Change title based on selected value
+      dynamicTitle: true,
+    },
+  },
+};
+
+const preview: Preview = {
```
pyphilia, line 39: Add a comment that it will always add theme, translations, etc to all components' stories, right?

src/modules/landing/header/RightHeader.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+  isAuthenticated: boolean;
+};
+
+const meta: Meta<RightHeaderProps> = {
+  component: RightHeader,
+  args: {
+    isAuthenticated: true,
+    onChangeLang: fn(),
+  },
+  // tags: ['autodocs'],
```
pyphilia, line 22: Remove?

pyphilia (approved):
Huge work! Thanks for the explanations, it looks really clean 💯

.storybook/preview.tsx
spaenleh: Yes this is correct, I will add a comment.

PR #556, opened 2024-12-17
fix: analytics test

spaenleh:
In this PR:
- move some files around
- fix cypress tests for analytics
- fix code smells


cypress/e2e/analytics/chartsHeader.cy.ts
```diff
@@ -64,8 +148,9 @@ describe('Select platform view ', () => {
 describe('Select users', () => {
   beforeEach(() => {
     cy.setUpApi({});
+    setupIntercepts();
   });
-  it('values of user select should be maintained when navigating within different routes', () => {
+  it.only('values of user select should be maintained when navigating within different routes', () => {
```
pyphilia, line 153: Remove.

src/locales/en/account.json
```diff
@@ -10,6 +10,7 @@
   "CROP_IMAGE_MODAL_UNEXPECTED_ERROR": "An unexpected error occurred when trying to crop the image, please try again later",
   "MAIN_MENU": {
     "HOME": "Home",
+    "STATS": "Statistics",
```
pyphilia, line 13: Maybe we want a clearer key?

src/modules/analytics/charts/TotalActionsByVerbChart.tsx
```diff
@@ -0,0 +1,122 @@
...
+    const formattedAggregateData = aggregateData.map((d) => ({
+      actionCount: (d.aggregateResult / totalActions) * 100,
+      type: d.actionType,
+    }));
+
+    // formattedAggregateData.forEach((d) => {
+    //   d.actionCount = parseFloat(
+    //     ((d.actionCount / totalActions) * 100).toFixed(2),
+    //   );
+    // });
```
pyphilia, line 74: Remove?

pyphilia (approved):
LGTM! 📶

src/locales/en/account.json
spaenleh: Do you mean change to
```suggestion
    "STATISTICS": "Statistics",
```

or 
```suggestion
    "STATISTICS_MAIN_MENU": "Statistics",
```
but it is already contextual no ? So I think short key is ok. I can use `STATISTICS` if you prefer.

PR #569, opened 2024-12-20
feat: implement tutorials page

pyphilia:
Current state: I added real tutorials, but didn't translate everything. I'll let you check them first. I didn't find any (not advanced) tutorial on player, so I removed the category.

The texts are full of bullshit from my pov, so we might want to work on it a bit more.


src/modules/landing/support/GithubLogo.tsx
spaenleh, line 1: The github logo already exists in the `~landing/footer/icons.tsx` file. I used it instead and removed this file

spaenleh (approved):
LGTM.

I pushed a few commits that mainly update the component code style to use export function instead of default exports and translated the nav bar.

PR #572, opened 2025-01-06
fix: add redirection page for login from legacy services

spaenleh:
In this PR I add 2 legacy supporting routes that support the usage of `/signin` and `/auth/signin` that is used by e.g. builder. They both redirect to the new `/auth/login` route. 

I have specified that these routes should be deleted in 6months.

### DIFF
--- a/src/routes/auth/signin.tsx
+++ b/src/routes/auth/signin.tsx
@@ -1,0 +1,10 @@
+import { createFileRoute, redirect } from '@tanstack/react-router';
+
+// legacy redirection for routes using the old `signin` route.
+// this should be removed 6 months after the push of the new client interface in production.
+// estimated date at time of writing is July 2025
+export const Route = createFileRoute('/auth/signin')({
+  beforeLoad: (ctx) => {
+    throw redirect({ to: '/auth/login', search: ctx.search });
+  },
+});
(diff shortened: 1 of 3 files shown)


src/routes/auth/signin.tsx
```diff
@@ -0,0 +1,10 @@
+import { createFileRoute, redirect } from '@tanstack/react-router';
+
+// legacy redirection for routes using the old `signin` route.
+// this should be removed 6 months after the push of the new client interface in production.
+// estimated date at time of writing is July 2025
```
pyphilia, line 5: I understand what you mean, but it's a little bit confusing when you read it, I think we can keep it simple as follow:
```suggestion
// this should be removed 6 months after the push of the new client interface in production, approximately around July 2025
```

src/routes/signin.tsx
```diff
@@ -0,0 +1,10 @@
+import { createFileRoute, redirect } from '@tanstack/react-router';
+
+// legacy redirection for routes using the old `signin` route.
+// this should be removed 6 months after the push of the new client interface in production.
+// estimated date at time of writing is July 2025
```
pyphilia, line 5: Same as above.

pyphilia (approved):

spaenleh:
@pyphilia Thank you for the re-phrasing indeed it was not very clear. Fixed.

PR #577, opened 2025-01-08
fix: link props with tanstack router update

spaenleh:
In this PR:
- update all packages to their latest version
- update tanstack/router link props given new version types


src/main.tsx
```diff
@@ -3,7 +3,6 @@ import ReactDOM from 'react-dom/client';
 import { HelmetProvider } from 'react-helmet-async';
 import { useTranslation } from 'react-i18next';
 import { ToastContainer } from 'react-toastify';
-import 'react-toastify/dist/ReactToastify.css';
```
spaenleh, line 6: In the newest version of `react-toastify` it is not necessary to include the css file manually.

spaenleh:
@pyphilia I changes the `enum` to a `const` and the "issue" is solved... let's see

pyphilia (approved):

PR #593, opened 2025-01-13
fix: social links to use react hook form

spaenleh:
In this PR I refactor the edit public profile component to use `react-hook-form` and simplify the logic.


src/modules/account/settings/publicProfile/EditPublicProfile.tsx
```diff
@@ -42,215 +34,158 @@ const config: Config = {
...
-        <CustomTextField
+      <Stack
+        component="form"
+        onSubmit={handleSubmit(onClose)}
+        direction="column"
+        gap={2}
+      >
+        <TextField
+          {...register('bio', {})}
           name="bio"
```
pyphilia, line 113: It's been a little since I used react hook forms, but I think you don't need to add the name manually anymore?

pyphilia (approved):
LGTM!

src/modules/account/settings/publicProfile/EditPublicProfile.tsx
spaenleh: Good catch, I forgot to remove it.

PR #596, opened 2025-01-13
fix: platform switch in analytics

spaenleh:
In this PR I add the platform switch back to analytics to allow switching to the player or builder view.

The analytics and library icons are disabled.

### DIFF
--- a/src/modules/analytics/PageWrapper.tsx
+++ b/src/modules/analytics/PageWrapper.tsx
@@ -16,7 +16,7 @@ import { Link, getRouteApi } from '@tanstack/react-router';
 
 import { UserSwitchWrapper } from '@/components/ui/UserSwitchWrapper';
 import { NS } from '@/config/constants';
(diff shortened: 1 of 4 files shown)


src/modules/analytics/PageWrapper.tsx
```diff
@@ -34,6 +39,22 @@ export function PageWrapper({
...
+  const platformProps = {
+    [Platform.Builder]: {
+      href: `${GRAASP_BUILDER_HOST}/items/${itemId}`,
+    },
+    [Platform.Player]: {
+      href: `/player/${itemId}/${itemId}`,
+    },
+    [Platform.Library]: {
+      disabled: true,
+    },
```
pyphilia, line 53: So it's impossible to go from analytics to library?

pyphilia (approved):
LGTM, but I had a question, we can discuss tomorrow 🔬

PR #606, opened 2025-01-14
fix: player home

### LINKED ISSUE #599: Update player home page to display a meaning full message
We are removing the home pages for the platforms and instead using the account home page to display the latest used items.

We should display a meaningful message on `/player` letting users know that they should choose an item to view on their profile

spaenleh:
In this PR I fix the player home by adding an automatic redirection to the account home.

Removed the tip card on account home.
Remove text translations fr UI that was removed (Guest limitation in player home)

close #599


pyphilia (approved):
LGTM ⭐

PR #729, opened 2025-01-23
fix: chatbox container

spaenleh:


src/config/env.ts
```diff
@@ -17,7 +17,7 @@ export const RECAPTCHA_SITE_KEY = import.meta.env.VITE_RECAPTCHA_SITE_KEY;
 
 export const H5P_INTEGRATION_URL =
   import.meta.env.VITE_GRAASP_H5P_INTEGRATION_URL ||
-  `${API_HOST}/p/h5p-integration`;
+  `${API_HOST}/items/h5p-assets/integration.html`;
```
spaenleh, line 20: This is the updated route in the backend to serve H5P in dev.

PR #740, opened 2025-01-24
fix: public profile visibility not updating

spaenleh:
In this PR I fix 2 bugs related to the public profile setting in account.

- ~~When the user does not have a public profile, the skeleton is shown even when the public profile has been fetched and returned a 204. Fixed in the linked PR:~~
  - https://github.com/graasp/graasp-query-client/pull/1020
- The visibility checkbox was not updating correctly react-hook-form because the ref was not set on the `FormControl`. I also forgot to add a watch on the value.

These fixes should improve the public profile editing experience.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "@emotion/react": "11.14.0",
     "@emotion/styled": "11.14.0",
     "@fontsource-variable/nunito": "5.1.1",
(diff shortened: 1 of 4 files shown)


spaenleh:
@pyphilia I reduced the scope of this PR.

PR #747, opened 2025-01-27
fix: react 19 migration

### LINKED ISSUE #746: Upgrade to react 19
We should upgrade to react 19.

It will be necessary to upgrade to the latest version of `react-leaflet`.
Migrating to react 19 will break `react-quill` since it uses an API which has been dropped in react 19: `findNodeById`. There is an up-to-date fork that exists `react-quill-new` which will work with react-19 and has even upgraded to the newer v2 of the quill.js package. From my research it will be a drop-in-replacement. 

Needed changes:
- JSX namespace is not global anymore, need to import from `react` package.
- refs have changed slightly, a workaround for the refs exported from react-dnd is needed: https://github.com/react-dnd/react-dnd/issues/3655

These changes can be done in bulk by using the provided codemods: `pnpm dlx types-react-codemod preset-19 ./src` Ref: https://github.com/eps1lon/types-react-codemod#scoped-jsx

spaenleh:
In this PR:

- Migrate to react 19
- Switch to react-quill-new from react-quill since it supports react19
- run the codemod to fix the JSX namespace changes
- fix some refs issues
- fix some types from `ReactElement<any>` to `ReactNode` which is better in a lot of cases.

close #746


src/components/chatbox/context/CurrentMemberContext.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { ReactNode } from 'react';
+import React, { type JSX, ReactNode } from 'react';
```
pyphilia, line 1: I think you can also use `type` for `ReactNode`? 
Remove `React` since you're changing imports?

src/modules/analytics/charts/ActionsMap.tsx
```diff
@@ -35,7 +35,8 @@ type BBox =
 const ActionsMap = (): JSX.Element | null => {
   const theme = useTheme();
   const { t } = useTranslation(NS.Analytics);
-  const mapRef = useRef<any>();
+  // TODO: replace with a better type, maybe by replacing the google map with leaflet
+  const mapRef = useRef<any>(undefined);
```
pyphilia, line 39: Factor out in an issue?

src/ui/Card/Card.tsx
```diff
@@ -54,7 +54,7 @@ type CardProps = {
    * image link to display as thumbnail
    */
   thumbnail?: string;
-  footer?: string | ReactElement;
+  footer?: string | ReactNode;
```
pyphilia, line 57:
```suggestion
  footer?: ReactNode;
```

pyphilia (approved):
Huge 🐘

src/modules/analytics/charts/ActionsMap.tsx
spaenleh: Done: https://github.com/graasp/client/issues/749

PR #763, opened 2025-01-28
fix: default to builder when user is already authenticated

spaenleh:
When user is authenticated, default to redirecting to the builder page if there is not url present.


src/routes/analytics/index.tsx
```diff
@@ -1,22 +1,47 @@
...
+          alignItems="center"
+          gap={6}
+          p={6}
+        >
+          <ChartNoAxesCombinedIcon size={100} />
+          <Typography variant="h4" textAlign="center">
+            {t('NO_ITEM_SELECTED')}
+          </Typography>
+        </Stack>
+        <ButtonLink to="/account">Choose an item</ButtonLink>
```
pyphilia, line 43: Translate?

pyphilia (approved):
Thanks for this PR! 🩵

src/routes/analytics/index.tsx
spaenleh: Good catch !

PR #818, opened 2025-01-31
fix: add sentry session for Release Health reporting

spaenleh:
In this PR I add the sentry release health feature by setting the current user when we know they are authenticated. This will help to track release adoption and health.
I only pass `id` and `name` as info to share the least possible.

### DIFF
--- a/src/AuthContext.tsx
+++ b/src/AuthContext.tsx
@@ -87,9 +87,7 @@ export function AuthProvider({
   }, [currentMember]);
 
   const value = useMemo(() => {
-    Sentry.setUser(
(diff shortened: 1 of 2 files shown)


src/AuthContext.tsx
```diff
@@ -82,6 +87,9 @@ export function AuthProvider({
   }, [currentMember]);
 
   const value = useMemo(() => {
+    Sentry.setUser(
+      currentMember ? { id: currentMember.id, name: currentMember.name } : null,
```
pyphilia, line 91: I think you can pass only the id?

pyphilia (approved):
Great! Thanks for the PR!

PR #822, opened 2025-02-03
fix: remove duplicated crop modal code

spaenleh:


src/modules/builder/components/thumbnails/ThumbnailCrop.tsx
```diff
@@ -22,7 +22,7 @@ import {
 
 import CropModal, {
   MODAL_TITLE_ARIA_LABEL_ID,
-} from '~builder/components/common/CropModal';
+} from '~account/home/memberPicture/CropModal';
```
pyphilia, line 25: Shouldn't the crop modal be in components, so that `builder` does not use `account`'s component?

pyphilia (approved):
LGTM, I've left a suggestion about the path the component should have.

PR #823, opened 2025-02-03
fix: flicker caused by suspense when loading translations

spaenleh:
While trying to fix the tests in #803 I had some weird test failures caused by what looked liked re-renders.

When investigting what could possibly cause these re-renders (when clicking to change the item type in the add new item modal), I saw that the translation file for the `messages` namespace was being loaded. 

Coupled with this insight I set on looking why loading new translations might cause a full page reload ... and I found https://stackoverflow.com/questions/68455594/page-flickers-when-lazy-loading-translations-with-i18next-http-backend

And indeed, adding the `{react: { useSuspense: false } }` setting in the i18n config solved the issue. 

I also changed the PlatformSwitch buttons to be TSR links instead of dumb anchor tags, which removes the flickering when navigating between account and builder using the navigation switch. Neat !

Pfiou finally found the reason behind the flickering issues. 💪

### DIFF
--- a/src/config/i18n.ts
+++ b/src/config/i18n.ts
@@ -25,4 +25,8 @@ i18n
       order: ['localStorage', 'navigator', 'querystring'],
       caches: ['localStorage'],
     },
+    react: {
+      // prevent the translations from using the react suspense and thus making the interface flicker
+      useSuspense: false,
+    },
   });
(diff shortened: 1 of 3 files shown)


pyphilia (approved):
💯

PR #839, opened 2025-02-04
fix: add home page and translation for page titles in common

spaenleh:
In this PR: 
- rename `/account` to `/home`
- add a wrapper to prevent guests to access the account pages, they will see a "Can't access message" and a back button if back is available
- update links to `/account` to use `/home`


src/routes/player/$rootId/$itemId.tsx
```diff
@@ -79,7 +79,13 @@ function PlayerWrapper(): JSX.Element {
 
   return (
     <Main
-      open={Boolean(rootId)}
+      open={
+        /**
+         * override the open prop to close the menu when there is no rootId or we could not fetch the item (logged out and not public)
+         * we want to keep the default behavior when the user is logged in
+         */
+        Boolean(rootId) && Boolean(item) ? undefined : false
```
spaenleh, line 87: Revert

PR #844, opened 2025-02-05
feat: add settings in member drop down

pyphilia:
Screenshot 2025-02-05 at 14 55 15

I added `settings` in the user drop down. 
The `UserSwitchWrapper` needs a lot of refactor, but I did a little bit while at it:
- factor out some translations
- show smaller icons

Remaining changes I can envision for further PRs:
- change the `redirect` mechanism to use links
- refactor the component to get actions as `children` instead as props. That would allow to factor out the `hooks.signOut`
- At some point get rid of the "wrapper"

### DIFF
--- a/src/ui/UserSwitch/UserSwitchWrapper.tsx
+++ b/src/ui/UserSwitch/UserSwitchWrapper.tsx
@@ -1,14 +1,18 @@
 import type { JSX } from 'react';
+import { useTranslation } from 'react-i18next';
 
 import {
   AccountCircle as AccountCircleIcon,
   MeetingRoom as MeetingRoomIcon,
 } from '@mui/icons-material';
-import { ListItemIcon, MenuItem, Typography } from '@mui/material';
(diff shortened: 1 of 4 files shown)


src/ui/UserSwitch/UserSwitchWrapper.tsx
```diff
@@ -1,14 +1,18 @@
...
 import {
   AccountCircle as AccountCircleIcon,
   MeetingRoom as MeetingRoomIcon,
 } from '@mui/icons-material';
-import { ListItemIcon, MenuItem, Typography } from '@mui/material';
+import { Divider, ListItemIcon, MenuItem, Typography } from '@mui/material';
 
 import { AccountType, CurrentAccount, redirect } from '@graasp/sdk';
 
+import { Settings } from 'lucide-react';
```
spaenleh, line 12: Usually I prefer to import the icons with the `Icon` suffic, makes it easier to spot in the JSX.

```ts
import { SettingsIcon } from 'lucide-react';
```

spaenleh (approved):

PR #853, opened 2025-02-06
feat: add accessible items on home page

### LINKED ISSUE #833: add accessible items in the home page

pyphilia:
- Reduced the number of shown recent items
- display as is the accessible items (no map)
- add same tests as builder "home"

Titles do not really make sense currently, because it is set lower in the components, and is used by pages like "published" or "trash". Should we decrease all titles, or pass down the titles are components?

https://github.com/user-attachments/assets/c022961b-6e57-4fe6-b985-ccfa87e44eb5

close #833

### DIFF
--- a/cypress/e2e/home/home.cy.ts
+++ b/cypress/e2e/home/home.cy.ts
@@ -1,0 +1,156 @@
+import {
+  ACCESSIBLE_ITEMS_ONLY_ME_ID,
+  CREATE_ITEM_BUTTON_ID,
+  HOME_LOAD_MORE_BUTTON_SELECTOR,
+  ITEM_SEARCH_INPUT_ID,
+  SORTING_ORDERING_SELECTOR_ASC,
+  SORTING_ORDERING_SELECTOR_DESC,
+  SORTING_SELECT_SELECTOR,
+  buildItemCard,
+} from '../../../src/config/selectors';
+import { SortingOptions } from '../../../src/modules/builder/components/table/types';
+import { ITEM_PAGE_SIZE } from '../../../src/modules/builder/constants';
+import { CURRENT_MEMBER } from '../../fixtures/members';
+import { ItemForTest } from '../../support/types';
+import { generateOwnItems } from '../builder/fixtures/items';
+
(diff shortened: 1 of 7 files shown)


cypress/e2e/home/home.cy.ts
```diff
@@ -0,0 +1,156 @@
...
+} from '../../../src/config/selectors';
+import { SortingOptions } from '../../../src/modules/builder/components/table/types';
+import { ITEM_PAGE_SIZE } from '../../../src/modules/builder/constants';
+import { CURRENT_MEMBER } from '../../fixtures/members';
+import { ItemForTest } from '../../support/types';
+import { generateOwnItems } from '../builder/fixtures/items';
+
+const HOME_PATH = '/home';
+
+const ownItems = generateOwnItems(30);
```
spaenleh, line 19: This fixture is not used in when setting up the api, but is used in a test, is this intentional ? maybe add a comment for why this is.

src/routes/_memberOnly/home.tsx
```diff
@@ -1,20 +1,66 @@
...
+
+  const itemSearch = useItemSearch();
+
   return (
-    <Stack gap={4} alignItems="center">
-      <MemberCard />
-      <Divider flexItem />
+    <Stack>
+      <Typography variant="h2" component="h1">
+        My Home
```
spaenleh, line 37: Translate ?

src/routes/_memberOnly/home.tsx
```diff
@@ -1,20 +1,66 @@
...
+        title={t('MY_GRAASP')}
+        options={
+          <Stack
+            direction="row"
+            alignItems="center"
+            justifyContent="flex-end"
+            spacing={1}
+          >
+            {itemSearch.input}
+            <NewFolderButton type={isMd ? 'button' : 'icon'} />
```
spaenleh, line 51: We should likely use container queries for these, so they are intelligent and they can decide for themselves how to render based on the container available space.

spaenleh (approved):

PR #855, opened 2025-02-06
feat: add menu tabs, remove drawer from home

pyphilia:
- Add `MenuTab` component
- Add Home, Published, Trash tabs to `/home`
- Remove drawer from `_memberOnly` 
- Create `home` namespace for traductions

Image

Planned for next PRs:
- Add tabs to `/account/settings` and `/account/storage`
- Move `/builder/_layout/trash` to`/home/_layout/trash`
- Move `/builder/published` to`/home/published`


src/modules/home/MenuTabs.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { useRouterState } from '@tanstack/react-router';
+import { HomeIcon, LibraryBigIcon, TrashIcon } from 'lucide-react';
+
+import { NS } from '@/config/constants';
+import { MenuTab } from '@/ui/MenuTab/MenuTab';
+
+export function MenuTabs() {
+  const {
+    location: { pathname },
+  } = useRouterState();
```
spaenleh, line 14: Actually tanstack router has support for changing the props of a component based on whether the target of the link is the current page, it needs to be implemented as a layout. But this means you do not need to implement the logic yourself. We can discuss this in more detail if you want me to show you how it can be done.

src/ui/MenuTab/MenuTab.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Stack, styled } from '@mui/material';
+
+import { Link } from '@tanstack/react-router';
+
+import { DEFAULT_LIGHT_PRIMARY_COLOR } from '../theme';
+
+const StyledLink = styled(Link)(() => ({
+  textDecoration: 'none',
+  color: 'inherit',
+}));
```
spaenleh, line 12: In order for Tanstack to be able to correctly type check the destination of the links, as well as do pre-fetching and such, the integration should be done like in the following docs article: https://tanstack.com/router/latest/docs/framework/react/guide/custom-link#mui-example

There are already some examples in the `@/components/ui` folder if you want to have a look !

spaenleh (approved):
Very interesting ! I like how you experimented with the tabs on the top. Maybe this would be best implemented with a layout and nested pages for the content that needs to change ? Having different routes will also allow to make fetching the data more focused and thus easier once we add loader functions (tanstack route loaders) for SSR.

PR #871, opened 2025-02-10
feat: add bookmarks in home

spaenleh:
Screenshot 2025-02-10 at 16 29 03


src/modules/home/bookmarks/BookmarkedItems.tsx
```diff
@@ -0,0 +1,70 @@
...
+  <Grid size={{ xs: 12, sm: 4, md: 3, xl: 2 }}>{children}</Grid>
+);
+const placeholderItems = [
+  { id: '6a704d93-3f00-4ff6-a142-77bf4192aef6' },
+  { id: '8507cea8-a95f-4650-8cdd-d4065488f1dc' },
+  { id: 'dbb1a033-c056-4753-8fd4-a4691581f3ad' },
+  { id: '8d8ad96d-a7b5-40d9-9a51-1472a126e35e' },
+  { id: '9858060a-0a22-457e-9f64-0d0f61360a0a' },
+  { id: '652fa7f0-1653-4baf-8a00-e533dafc6655' },
+];
```
pyphilia, line 20: Do you really need those ids? I can guess it's mock data, but it's a bit strange to find those randomly set ids in production code. Wouldn't it be simpler to use the `index` and disable the eslint rule?

src/modules/home/bookmarks/BookmarkedItems.tsx
```diff
@@ -0,0 +1,70 @@
...
+          </GridWrapper>
+        ))}
+      </Grid>
+    );
+  }
+
+  if (isError) {
+    return (
+      <Alert severity="error">
+        There was an issue retrieving your favorite items.
```
pyphilia, line 64:
```suggestion
        There was an issue retrieving your bookmarked items.
```
translate?

src/modules/home/bookmarks/BookmarkedItems.tsx
```diff
@@ -0,0 +1,70 @@
...
+      </Alert>
+    );
+  }
+
+  // else we have nothing to show
```
pyphilia, line 69: Should we return `null` here?

pyphilia (approved):
Thanks for the PR! I've left some suggestions, but LGTM 🦀

src/modules/home/bookmarks/BookmarkedItems.tsx
spaenleh: Using the index is prone to errors if the list was to be modified. I can add a comment, but I think it is better to have specific ID (for placeholder data).

PR #906, opened 2025-02-12
fix: update poper user menu

### LINKED ISSUE #835: Improve the user dropdown by making it more obvious it can be clicked
Possible change could be to remove the name display and use only the avatar picture.

This would make it also more compact.

### LINKED ISSUE #870: Notification icon shows in bar even when member is not logged in

spaenleh:
close #835

Screenshot 2025-02-12 at 12 45 56

also fix #870 by including the mentions icon inside the header right component only when user is authenticated.


src/locales/en/common.json
```diff
@@ -64,7 +64,8 @@
     "PROFILE": "Profile",
     "SETTINGS": "Settings",
     "STORAGE": "Storage",
-    "SIGN_IN": "Sign In",
+    "LOG_IN": "Log In",
+    "LOG_OUT": "Log out",
```
pyphilia, line 68: Majuscule? One word? Maybe uniformize?

src/ui/UserPopupMenu.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+    // sign out button
+    expect(menuCanvas.getByText(args.signOutText)).toBeInTheDocument();
+  },
+} satisfies Story;
+
+export const Guest = {
+  args: {
+    signOutText: 'Sign Out',
+    currentMember: { ...CURRENT_MEMBER, type: AccountType.Guest },
```
pyphilia, line 68: use `GuestFactory`? You can define fixed properties given your test.
Same for above `MemberFactory`?

src/ui/UserPopupMenu.tsx
```diff
@@ -0,0 +1,131 @@
...
+  };
+
+  const open = Boolean(anchorEl);
+  return (
+    <>
+      <IconButton
+        onClick={handleClick}
+        size="small"
+        id={buttonId}
+        aria-controls={open ? 'account-menu' : undefined}
```
pyphilia, line 76: What is it related for? I cannot find the reference.

pyphilia (approved):
📑

src/ui/UserPopupMenu.stories.tsx
spaenleh: Actually they are not full members, they come from the authContext where I create a `AuthenticatedMember` which has the following shape:
```ts
type AutenticatedMember = {
  id: string;
  name: string;
  lang: string;
  type: 'individual' | 'guest';
}
```

src/ui/UserPopupMenu.tsx
spaenleh: I copied it from the mUI docs, but it should be the id of the Menu (poper). I have updated it to be correct.

src/locales/en/common.json
spaenleh: I use two words because it is the verb not the substantive (noun). I will make them both lowercase.

PR #916, opened 2025-02-13
fix: add limit on bookmarks shown

### LINKED ISSUE #877: Limit number of bookmarks shown on home
We should limit the number of bookmarks shown on home, so it does not take the whole view. 

Also propose a way to see all your bookmarks in a single page.

spaenleh:
close #877

### DIFF
--- a/src/modules/home/bookmarks/ItemFavoriteCard.tsx
+++ b/src/modules/home/bookmarks/ItemFavoriteCard.tsx
@@ -19,7 +19,7 @@ type Props = {
   item: PackedItem;
 };
 
-export function ItemCard({ item }: Readonly<Props>): JSX.Element {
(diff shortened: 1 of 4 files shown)


src/modules/home/bookmarks/ItemFavoriteCard.tsx
```diff
@@ -19,7 +19,7 @@ type Props = {
   item: PackedItem;
 };
 
-export function ItemCard({ item }: Readonly<Props>): JSX.Element {
+export function ItemFavoriteCard({ item }: Readonly<Props>): JSX.Element {
```
pyphilia, line 22:
```suggestion
export function BookmarkCard({ item }: Readonly<Props>): JSX.Element {
```

pyphilia (approved):
Great! I just suggested to rename a component^^

PR #918, opened 2025-02-13
fix: change profile to home in user menu popup

### LINKED ISSUE #914: Change "profile" text in poper to "home"

spaenleh:
In this PR:
Change text in user menu popup from "Profile" to "Home".

In builder:
Screenshot 2025-02-13 at 15 47 09

In home (notice that since we are in the /home route, the item is selected):
Screenshot 2025-02-13 at 15 47 02

close #914


pyphilia (approved):
🧑‍🍳

PR #919, opened 2025-02-13
fix: refactor menuTabs to use tsr link

### LINKED ISSUE #864: Refactor MenuTab component into CustomLink from TSR
In order for Tanstack to be able to correctly type check the destination of the links, as well as do pre-fetching and such, the integration should be done like in the following docs article: https://tanstack.com/router/latest/docs/framework/react/guide/custom-link#mui-example

There are already some examples in the `@/components/ui` folder if you want to have a look !

_Originally posted by @spaenleh in https://github.com/graasp/client/pull/855#discussion_r1948831155_

spaenleh:
close #864

### DIFF
--- a/src/modules/home/MenuTabs.tsx
+++ b/src/modules/home/MenuTabs.tsx
@@ -2,16 +2,12 @@ import { useTranslation } from 'react-i18next';
 
 import { Stack } from '@mui/material';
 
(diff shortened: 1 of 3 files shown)


pyphilia (approved):
Thanks for the changes!

PR #920, opened 2025-02-14
fix: make nested pages for home

### LINKED ISSUE #867: Make "published" and "trash" nested routes inside the home layout

spaenleh:
In this PR:
- make published and trash nested pages of the home wrapper
- remove the builder page layout and provide page title using staticData

nested-home-pages

Open Questions:
- What do we do with the "Bookmarks" section in the builder ? Is it enough to have them in the home ? 
- Should we keep the mode toggle in the home ? Should we keep the map view in home ? should it be a different button ?

close #867


src/modules/builder/components/pages/home/HomeScreenContent.tsx
```diff
@@ -81,16 +78,16 @@ export function HomeScreenContent({
...
+  // if (mode === ItemLayoutMode.Map) {
+  //   return (
+  //     <>
+  //       <Stack direction="row" justifyContent="flex-end">
+  //         <ModeButton />
+  //       </Stack>
+  //       <DesktopMap />
+  //     </>
+  //   );
+  // }
```
pyphilia, line 90: I think we can remove.

pyphilia (approved):

PR #924, opened 2025-02-14
fix: add umami data on bookmark card

### LINKED ISSUE #921: Track bookmark was clicked

spaenleh:
close #921 

Add umami event when clicking bookmark card


pyphilia (approved):
🐔

PR #925, opened 2025-02-14
fix: update validation for text on description change

### LINKED ISSUE #905: Trying to save an empty text does not allow to save a text anymore even with content
## Steps
- Open new item modal, select text
- Add a name
- Click on save button, expected behavior is that the save button is disabled and "cannot save empty error" is displayed
- fill content

Expected behavior: save button is enabled, error is hidden
Current behavior: the save button is not enabled, the error is still active

pyphilia:
close #905


spaenleh (approved):
Good catch ! I find it hard to use ReactQuill with react-hook-form. It works better out-of-the box with native html elements.

PR #938, opened 2025-02-19
fix: simplify ws config in query client, define a ws client for prod

pyphilia:
No ws client was defined for prod. I simplified the code as well to remove the `WS_HOST` prop.


spaenleh (approved):

spaenleh:
Thank you for the simplification !

PR #939, opened 2025-02-19
fix: app resizing issue

spaenleh:
In this PR I fix an issue with app resizing.

The current situation is that apps that want to be perfectly fit (using the autoresize hook) will send their height and it will be applied to the `height="xxx"` prop of the iframe. 
The apps that do not send this info will not have a height set, sot hey will be very small.

We want:
- apps using auto-resize should have their specified height used
- apps **not** using auto-resize should have a large height (around the full-viewport)
- apps using the resizable wrapper should be able to be resized by the component

The issue previously was that if we specified a height in the styling class, then the hook-provided height applied on the element directly would be override it, because it has lower specificity.

The fix is to change where we put the hook-provided height: using `ref.current.style.height` instead of `ref.current.height` we get the highest specificity possible, so it correctly overrides the default height.
We are then able to give the iframe a default height on the iframe class and all works well.

### DIFF
--- a/src/modules/builder/components/item/ItemContent.tsx
+++ b/src/modules/builder/components/item/ItemContent.tsx
@@ -140,7 +140,7 @@ const AppContent = ({
   <AppItem
     isResizable={false}
     item={item}
-    height={ITEM_DEFAULT_HEIGHT}
+    height="100%"
     requestApiAccessToken={(payload: {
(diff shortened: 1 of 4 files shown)


pyphilia (approved):
LGTM! 🎏

Issue #819, opened 2025-01-31
Notifier has not translation


spaenleh:
The toasts from the notifier are not translated correctly.

PR #940, opened 2025-02-19
fix: use update password generation to fix translation

pyphilia:
I replace the `updatePassword` and `createPassword` endpoints to avoid using `notifier`.

ref #819

### DIFF
--- a/src/modules/auth/components/signIn/MagicLinkLoginForm.tsx
+++ b/src/modules/auth/components/signIn/MagicLinkLoginForm.tsx
@@ -7,6 +7,7 @@ import { Stack } from '@mui/material';
 import { RecaptchaAction } from '@graasp/sdk';
 
 import { useLocation, useNavigate } from '@tanstack/react-router';
+import isEmail from 'validator/lib/isEmail';
 
 import { NS } from '@/config/constants';
 import { mutations } from '@/config/queryClient';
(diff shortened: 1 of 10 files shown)


src/modules/auth/components/signIn/MagicLinkLoginForm.tsx
```diff
@@ -7,6 +7,7 @@ import { Stack } from '@mui/material';
 import { RecaptchaAction } from '@graasp/sdk';
 
 import { useLocation, useNavigate } from '@tanstack/react-router';
+import isEmail from 'validator/lib/isEmail';
```
spaenleh, line 10: I think we have one in sdk ? Is it not ok to use the isEmailValid that we have from auth ?

src/modules/auth/validation.ts
```diff
@@ -42,13 +42,6 @@ export const emailValidator = (email?: string) => {
   return isEmail(email, {}) ? null : INVALID_EMAIL_ERROR;
 };
 
-export const isEmailValid = (email?: string) => {
```
spaenleh, line 45: Ah you removed it ? should we have moved it where it is used instead ?

spaenleh (changes requested):
Changes look good, I am just not sure why you removed the email validation function to use the one from the validator package ? I tried to remove the dependency on that package by copying only the validation function we used into sdk.

src/modules/auth/validation.ts
pyphilia: It was indeed using some indirection I removed. I changed for having the same mechanism as in registration, using `isEmail` from sdk.

PR #941, opened 2025-02-19
feat: use export member data generation, remove notification

pyphilia:
I removed the `exportData` mutation to use the generated one. I removed the use of the notifier, showing alert instead.


spaenleh (approved):
LGTM ! 💮

PR #946, opened 2025-02-27
feat: add setting to allow reader to edit etherpad

pyphilia:
TODO
- [ ] check on dev

### DIFF
--- a/src/modules/builder/components/item/form/etherpad/EtherpadForm.tsx
+++ b/src/modules/builder/components/item/form/etherpad/EtherpadForm.tsx
@@ -10,19 +10,23 @@ import {
   Typography,
 } from '@mui/material';
 
-import { DiscriminatedItem } from '@graasp/sdk';
+import { DiscriminatedItem, PermissionLevel } from '@graasp/sdk';
+
+import { useMutation, useQueryClient } from '@tanstack/react-query';
 
 import { NS } from '@/config/constants';
-import { mutations } from '@/config/queryClient';
 import { ITEM_FORM_CONFIRM_BUTTON_ID } from '@/config/selectors';
+import { createEtherpadMutation } from '@/openapi/client/@tanstack/react-query.gen';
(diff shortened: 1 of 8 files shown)


src/modules/builder/components/item/form/etherpad/EtherpadForm.tsx
```diff
@@ -41,7 +52,17 @@ export function EtherpadForm({
...
   const onSubmit = async (data: Inputs) => {
-    await postEtherpad({ parentId, name: data.name });
+    await postEtherpad({
+      query: {
+        parentId,
+      },
+      body: {
+        name: data.name,
+        readerPermission: data.allowReadersToWrite
+          ? PermissionLevel.Write
```
spaenleh, line 62: Should these be using the EtherpadPermission enum instead ?

src/modules/player/item/Item.tsx
```diff
@@ -89,7 +89,8 @@ type EtherpadContentProps = {
 const EtherpadContent = ({ item }: EtherpadContentProps) => {
   const { t } = useTranslation(NS.Common);
   // get etherpad url if type is etherpad
-  const etherpadQuery = useEtherpad(item, 'read');
+  // server will return read view if no write access allowed
+  const etherpadQuery = useEtherpad(item, 'write');
```
spaenleh, line 93: In this case do we need to pass the "write" argument ? It looks the backend decides if you get a read or a write anyways... it just looks confusing to specify write here I would say.

spaenleh (approved):
Changes look good! I have left some minor comments. Maybe we can tackle them in other PR after this one to improve the readability of the useEtherpad hook.

src/modules/player/item/Item.tsx
pyphilia: The original idea is that you can disable the write access from the frontend, eg. preview mode. By default you get `read` mode unless you specify.

PR #953, opened 2025-03-03
feat: allow to delete bookmarks on home

### LINKED ISSUE #922: Add bookmarks management (title, delete)
- Add bookmarks title with icon
- Add manage button (to allow deleting)

pyphilia:
https://github.com/user-attachments/assets/fa4a8cfa-491e-4215-907c-3989356da143

close #922

### DIFF
--- a/src/locales/en/home.json
+++ b/src/locales/en/home.json
@@ -8,7 +8,10 @@
   },
   "BOOKMARKED_ITEMS": {
     "SHOW_LESS": "Show less",
-    "SHOW_MORE": "Show more"
+    "SHOW_MORE": "Show more",
+    "MANAGE_BUTTON": "Manage",
+    "CANCEL_BUTTON": "Back",
+    "TITLE": "Bookmarks"
   },
   "LIBRARY": {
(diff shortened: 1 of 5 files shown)


spaenleh:
> [!NOTE]
> I haven't looked at the code yet, so this is just a comment on aesthetics from watching the recording.

Since the only actions that the user can perform in this new "modify" view is to remove bookmarks, maybe the removal button should more clearly communicate that it will remove the whole card by placing it on the corner of the card.

Also I think the choice of the "X" icon and the "back" text did not resonate with me, I was not sure what they would do. 

Below you can find a screenshot with suggestions that I think could improve the the understanding of the feature by basing they on common interactions in the digital world. Let me know what you think. 

I like the choice of red for the deletion icon, and the rest of the interaction and button placement is 👌 .

suggestion

pyphilia:
Screenshot 2025-04-28 at 10 59 43
I updated with this design

spaenleh (approved):
📖 Looks good ! Thank you for this update !

PR #958, opened 2025-03-04
fix: redirect builder to home

pyphilia:
Prevent `/builder` to lead to a non found route. Instead it leads to `/home`, our new home page.


MatveyK (approved):
LGTM.

PR #962, opened 2025-03-18
fix: use language query string when present

spaenleh:
In this PR I update the language detection priority so the language query string is considered before the local storage or the navigator.

This allows urls from the documentation to specify for example: `http://graasp.org/auth/register?lang=fr` and the register page will be in french.

### DIFF
--- a/src/config/i18n.ts
+++ b/src/config/i18n.ts
@@ -23,9 +23,14 @@ i18n
     supportedLngs: ['fr', 'de', 'it', 'es', 'ar', 'en'],
     // options for the language detector
     detection: {
+      // The priority order is defined as:
+      // 1. querystring: users coming from an external link may want to see the page in the language set by the link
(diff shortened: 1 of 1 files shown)


src/config/i18n.ts
```diff
@@ -23,8 +23,10 @@ i18n
     supportedLngs: ['fr', 'de', 'it', 'es', 'ar', 'en'],
     // options for the language detector
     detection: {
-      order: ['localStorage', 'navigator', 'querystring'],
+      order: ['querystring', 'localStorage', 'navigator'],
```
pyphilia, line 26: Maybe we could explain why query string comes first?

pyphilia (approved):

src/config/i18n.ts
spaenleh: Yes I added some comments to describe our choices.

PR #999, opened 2025-05-09
fix: change membership related hooks and mutations

pyphilia:


src/modules/builder/components/item/sharing/membershipTable/EditPermissionButton.tsx
```diff
@@ -1,7 +1,9 @@
 import { type JSX, useState } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 
+import { LoadingButton } from '@mui/lab';
```
spaenleh, line 4: This is now built into the `Button` api I think. https://v6.mui.com/material-ui/react-button/#loading-2

src/modules/builder/components/item/sharing/membershipTable/EditPermissionButton.tsx
```diff
@@ -92,21 +98,28 @@ const EditPermissionButton = ({
...
-              handleUpdate(currentPermission);
-              closeModal();
+            onClick={async () => {
+              try {
+                await handleUpdate(currentPermission);
+                closeModal();
+              } catch (e) {
+                console.error(e);
+                setError(translateMessage(getErrorMessage(e)));
+              }
```
spaenleh, line 118: Should this be extracted as a function ? looks like a lot of code to put inline.

src/modules/builder/components/item/sharing/membershipTable/useMembershipRequests.ts
```diff
@@ -0,0 +1,60 @@
...
+import { useMutation, useQueryClient } from '@tanstack/react-query';
+
+import { hooks } from '@/config/queryClient';
+import {
+  createItemMembershipMutation,
+  deleteMembershipRequestMutation,
+} from '@/openapi/client/@tanstack/react-query.gen';
+import { itemKeys } from '@/query/keys';
+
+export const useMembershipRequests = ({
```
spaenleh, line 12: use `function` to be consistent ?

spaenleh (changes requested):
Code Changes look good, nice to see some of the custom hooks replaced by the generation. 

I just have some questions related to the regexes for the server mocks.

cypress/support/server.ts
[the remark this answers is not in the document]
pyphilia: It's because of the generation, the backend register the root to `/`, which happens automatically a `/` at the end of the route. It does not matter when running the app, but for test with cypress, the intercepts need this `/`.

spaenleh (approved):
Looks good, thank you for the changes. Also it makes sens about the `/` at the end, I did not know that it was possible. As discussed I think it is more common to not have them at the end in the backend. Lets merge !

PR #1000, opened 2025-05-14
fix: add docker deploy

spaenleh:
In this PR I add an option to deploy to docker.

There a few limitations to deploying with docker:
- we need to supply all the env vars at build time
- I had to rename `VITE_RECAPTCHA_SITE_KEY` to `VITE_RECAPTCHA` because docker complained that I was exposing potential credentials


.dockerignore
```diff
@@ -0,0 +1,13 @@
+.github
+.nyc_output
+.storybook
+.vscode
+build
+coverage
+docs
+node_modules
+storybook_static
+.env*
```
pyphilia, line 10: I think it started alphebetically, then you added some other files. Add a comment? Sort again?

pyphilia (approved):
Thanks for this PR, but what was this needed for?

spaenleh:
@pyphilia thank you for your review.

This makes it possible to run the full Graasp platform from a single docker-compose file, easing the cost to entry for people wanting to run their own instance of graasp at home. 

However it is not perfect, since the client is essentially statically built assets and we need to configure them when building, so users are still require to build the image themselves, which is not the easiest thing to do. I try to include instructions in the readme. 

If we wanted to remove the need for env variables we could:
- expect the api to be available on the same origin under the `/api` route prefix
- get the recpatcha site key from an api call to the backend
- hardcode the h5p route in the backend

These changes would make it possible to build the client and publish it without leaking sensitive keys or making it impossible to use on another origin (since it would be expected that the api runs on the same origin).

PR #1025, opened 2025-06-23
refactor: use bookmark generated client, remove manual bookmark

pyphilia:
Remove bookmarks manual hooks and mutations to use the openapi-generated ones.

The generation gives us `PackedItem`, which conflicts with `DiscriminatedItem`. For simplicity I did 2-3 casts for it to work with the current `graasp/sdk`.


spaenleh (approved):
Nice, for the casts, I do not know, maybe there is something we need to do in the api ? but otherwise we should probably change the helper functions to work better with the generated types ...

PR #1029, opened 2025-06-24
fix: migrate hey-api to remove `@hey-api/client-fetch` package

spaenleh:


package.json
```diff
@@ -21,11 +21,11 @@
     "@fontsource-variable/nunito": "5.1.1",
     "@graasp/sdk": "5.13.3",
     "@graasp/stylis-plugin-rtl": "2.2.0",
-    "@hey-api/client-fetch": "0.10.0",
-    "@sentry/react": "9.31.0",
+    "@hey-api/openapi-ts": "0.76.0",
```
pyphilia, line 24: Should this be in devDependencies?

pyphilia (approved):

package.json
spaenleh: @pyphilia I was able to verify that indeed the package can be added as a devDependency since it is used only for code-gen, the build ouput does not require the package anymore! very cool !

PR #1031, opened 2025-06-24
refactor(chatbox): use generated client for chatbox

pyphilia:
This PR refactors the Chatbox to use the generated endpoints. I removed some props and context to rather use their corresponding generated hooks. I remove references to the types from `@graasp/sdk` (ie. `ChatMessageWithCreator`) and rather use the generated one.

I created a `useChatboxProvider` hook to encapsulate the mutations, but also the websocket setup. I believe it was kinda already the case, but setting up the websocket multiple times will subscribe multiple times, and thus we might receive the feedback multiple times (ie. sending one message can end up in receiving 4x the same message). This PR does not have a mechanism for handling such case, it assumes the chat will be displayed only once.

TODO:
- [x] write tests for `useChatboxProvider`


vitest.config.ts
```diff
@@ -9,8 +10,21 @@ export default defineConfig({
...
+        },
+      },
+      {
+        test: {
+          name: 'hooks',
+          include: ['src/**/*.hook.test.ts'],
           exclude: [...configDefaults.exclude, queryClientInclude],
+          environment: 'happy-dom',
         },
+        plugins: [tsconfigPaths()],
```
pyphilia, line 27: I wanted to write vitest test for my hook. However I got 2 problems:
- It didn't recognise aliases in paths
- It didn't have `document`

So I added a project for testing `hooks` (`*.hook.test.ts`) in `happy-dom` with a plugin to read alias. I excluded those in the classic `unit` tests.  

Let me know what you think of this. This might be useful for other hook test we would like to do.

spaenleh: I see, two things:
- you can add `extends: './vite.config.js',` in the project config so it inherits the vite config (which already contains the path alias and other plugins)
- Maybe we want to change the "query" project to also include "hooks" tests. For this we can add multiple patterns in the include config and exclude them in other tests.

What do you think ? otherwise if you think it is usefull to have more projects, one specific to the query side and others for the hook, for unit tests etc, then yes it is file to create a new project entry.

src/components/chatbox/Chatbox/useChatboxProvider.hook.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+/**
+ * React hook to subscribe to the updates of the given chat ID
+ * @param chatId The ID of the chat of which to observe updates
+ */
+export const useItemChatUpdates = (chatId: UUID | null) => {
+  const WS_CLIENT = configureWebsocketClient(
+    `${API_HOST.replace('http', 'ws')}/ws`,
+  );
```
spaenleh, line 73: Do we want to re-define the ws client inside the hook everytime ? should it be provided as an argument to the hook instead ?

src/modules/builder/components/main/DownloadButton.tsx
```diff
@@ -33,7 +33,7 @@ export const DownloadButton = ({
   const { mutate: downloadItem, isPending: isDownloading } = useMutation({
     ...downloadFileMutation(),
     onSuccess: (file) => {
-      const url = window.URL.createObjectURL(new Blob([file]));
+      const url = window.URL.createObjectURL(new Blob([file as Blob]));
```
spaenleh, line 36: Was this fixed already ? Make sure when you rebase.

spaenleh (approved):
Great work !

PR #1047, opened 2025-07-22
fix: add catch for message parsing errors from apps

### LINKED ISSUE #1050: Can't parse messages from apps that are not using JSON format
There are users that register "app" items for third party apps that **do** send postMessage but not in the JSON format. 

## Bug

The app component receives the postmessage and tries to parse it with the JSON parser. 
This fails since the message is not a JSON object.

## Solution

Add a try-catch block around the failable logic so that the exception is handled.

spaenleh:
- fix #1050

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -99,6 +99,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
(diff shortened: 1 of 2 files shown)


pyphilia (approved):
LGTM 🌼

PR #1051, opened 2025-07-23
fix: notify users when the description is too long

### LINKED ISSUE #1049: Description too long does not show an error message and users can't save the text
The description length limit is 5000 chars. When using text with links and other markdown it is easy to exceed this size, we should warn users that the description is too long so they know why it can't be saved.

spaenleh:
In this PR I fix #1049 

Currently users can enter a description that is longer than the allowed length (5000 chars in the backend).

I fix this problem by:
- adding an error message when the description is too long
- disabling the submit button so users can't save a description that is too long
- make `create` and `edit` folder forms validate the inputs `onChange` to increase the feedback to the user (previously they would only validate `onSubmit`

While working on this I noticed that the validation mode for the create and edit modals for the folder was set to `onSubmit`. I do not know if there was a reason behind it. @pyphilia do you know ?

Also I changed the `DescriptionForm` component to use `<Controller />` from react-hook-form to make validation more straight forward.

I also changed the size of the edit dialog to match the one of the create (`maxWidth: md`).

Bellow is a demo of the changes:

https://github.com/user-attachments/assets/b314f331-cb10-497d-864e-3c0c0f1c0c4b

### DIFF
--- a/cypress/e2e/builder/item/create/createFolder.cy.ts
+++ b/cypress/e2e/builder/item/create/createFolder.cy.ts
@@ -1,5 +1,6 @@
 import { PackedFolderItemFactory } from '@graasp/sdk';
 
+import { MAX_DESCRIPTION_LENGTH } from '../../../../../src/config/constants';
 import {
   FOLDER_FORM_DESCRIPTION_ID,
   ITEM_FORM_CONFIRM_BUTTON_ID,
@@ -44,8 +45,6 @@ describe('Create Folder', () => {
     cy.visit(HOME_PATH);
     createFolder({ name: ' ' }, { confirm: false });
 
-    // button is not disabled at beginning
-    // cy.get(`#${ITEM_FORM_CONFIRM_BUTTON_ID}`).click();
     cy.get(`#${ITEM_FORM_CONFIRM_BUTTON_ID}`).should('be.disabled');
   });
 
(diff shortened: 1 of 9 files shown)


cypress/e2e/builder/item/create/createFolder.cy.ts
```diff
@@ -44,10 +45,25 @@ describe('Create Folder', () => {
     createFolder({ name: ' ' }, { confirm: false });
 
     // button is not disabled at beginning
-    cy.get(`#${ITEM_FORM_CONFIRM_BUTTON_ID}`).click();
+    // cy.get(`#${ITEM_FORM_CONFIRM_BUTTON_ID}`).click();
```
pyphilia, line 48: remove?

src/locales/en/messages.json
```diff
@@ -156,5 +156,6 @@
   "CANNOT_ENROLL_FROZEN_ITEM_LOGIN_SCHEMA": "Cannot enroll in the targeted item because the subscription has been frozen by its owner",
   "NOT_MEMBER_OR_GUEST": "Your account is not registered as member or pseudonimized user",
   "INVALID_ITEM_NAME_PATTERN_ERROR": "The item's name is not valid. It should start and end with an alphabetical character.",
-  "INVALID_ITEM_NAME_MAX_LENGTH_ERROR": "The item's name is too long."
+  "INVALID_ITEM_NAME_MAX_LENGTH_ERROR": "The item's name is too long.",
+  "INVALID_ITEM_DESCRIPTION_MAX_LENGTH_ERROR": "Your description is too long, keep descriptions short. Use a text element inside your page for longer descriptions."
```
pyphilia, line 160:
```suggestion
  "INVALID_ITEM_DESCRIPTION_MAX_LENGTH_ERROR": "Your description is too long, descriptions should be short. If you want longer descriptions, use a text element inside your page."
```

src/modules/builder/components/item/form/description/DescriptionForm.tsx
```diff
@@ -1,36 +1,59 @@
...
 import { NS } from '@/config/constants';
 import TextEditor from '@/ui/TextEditor/TextEditor';
 
 export type DescriptionFormProps = {
   id?: string;
-  value?: string;
-  onChange: (v: string) => void;
 };
 
+export const MAX_DESCRIPTION_LENGTH = 5000;
```
pyphilia, line 14: Should this be in config? So you can reuse in your tests perhaps, instead of "5001"

cypress/e2e/builder/item/edit/editFolder.cy.ts
```diff
@@ -19,7 +19,7 @@ const EDITED_FIELDS = {
 };
 
 describe('Edit Folder', () => {
-  it('confirm with empty name', () => {
+  it('cannot make name empty', () => {
```
pyphilia, line 22:
```suggestion
  it('cannot save item with empty name', () => {
```

pyphilia (approved):
Looks good to me. Maybe check the other forms since it changes their descriptions, but I guess that it works if the tests pass.

PR #1054, opened 2025-07-23
feat: setup collaboration with yjs for pages

pyphilia:
Setup yjs for collaboration.

Shows current active users.

Do not save content (some content may be persisted because of the server state).

https://github.com/user-attachments/assets/79ae3ab7-d52a-4bfc-b178-7d10077253bc


spaenleh (approved):
Looks good !

PR #1057, opened 2025-07-25
fix: improve builder folder header section

### LINKED ISSUE #682: Wrapping of item title does not look right
The item title is wrapping at max 50% of the width of the page even when there is available space before the search bar.
Screenshot 2023-09-29 at 09 46 09

pyphilia:
Improve header section in the Builder, based on the Player's. 

- Show thumbnail if it exists
- Improvements for small screens added for both Builder and Player
- Show creator name instead of update date in both Builder and Player
- Search and Create buttons are below the description

Here a screenshot for smaller title, with a thumbnail
Screenshot 2025-07-25 at 12 13 04

Below are resizable interfaces with long titles, with and without thumbnail.

https://github.com/user-attachments/assets/3d411c41-65c9-430a-aed9-21074c915141

https://github.com/user-attachments/assets/8c476e76-9a13-45ff-b012-ba9ee3b3d72a

fix #682


src/components/SectionHeader.tsx
```diff
@@ -0,0 +1,59 @@
...
+              maxHeight={isSm ? '96px' : '100%'}
+              url={thumbnailSrc}
+              alt={item.name}
+              sx={{ borderRadius: 5 }}
+            />
+          </Stack>
+        )}
+        <Stack gap={1} width="100%">
+          <Typography
+            className={FOLDER_NAME_TITLE_CLASS}
```
spaenleh, line 41: Do we need to use className ? is it only for tests ?

spaenleh (approved):
🔥 Wow very nice !

PR #1061, opened 2025-07-29
feat: show capsule examples on landing page

pyphilia:
Display capsule example on the home page.
Open to design ideas. Show creator? Add tags like "interactive app", "physics"? 
Currently the links go to the player pages.

The idea would be to show different capsules depending on the languages. Let's show english capsules if we lack content. 

I added titles to the sections (For teachers), I think it was missing.

https://github.com/user-attachments/assets/9d621567-158f-47ce-90e0-27f007225d12


src/modules/landing/home/UserStory.tsx
```diff
@@ -79,44 +79,49 @@ export function UserStory({
...
   return (
-    <InvertingStack gap={5} mt={6} position="relative">
-      <Box id={id} sx={{ position: 'absolute', top: '-200px' }} />
-      <Stack className={IMAGE_CONTAINER} flex={1} gap={1} alignItems="center">
-        <Stack borderRadius={6} overflow="hidden">
-          {image}
+    <Stack>
+      <Typography color="primary" variant="h1" component="h2">
+        For Teachers
+      </Typography>
```
spaenleh, line 85: Make it a prop so it can be different for the teachers and the researchers section.

src/modules/landing/home/UserStorySection.tsx
```diff
@@ -21,22 +27,53 @@ export function UserStorySection(): JSX.Element {
...
+                    imageSrc={imageSrc}
+                    description={description}
+                    url={url}
+                  />
+                </Grid>
+              );
+            })}
+          </Grid>
+
+          <Button
```
spaenleh, line 67: Use `ButtonLink` ? with `href` prop instead of `to` for external links

spaenleh (approved):
I like the idea, it looks good and makes the home page a bit more lively !

pyphilia:
I changed the design, so we suppose Library will get a broad nice homepage!

Screenshot 2025-09-09 at 18 08 47

PR #1064, opened 2025-07-31
fix: use english names for builder, player, analytics

pyphilia:
Should we enforce for the other languages as well?


spaenleh (approved):

spaenleh:
I don't know for the other languages ... depends if we consider it to be a "trademark" or if we want to adapt to the local language.

PR #1071, opened 2025-08-04
fix: guests can not change their language

spaenleh:


pyphilia (approved):
LGTM, I suggested a better text for the button 🧊

PR #1081, opened 2025-08-21
fix(page): timeout connection attempts if disconnections are repeated

pyphilia:
In case the server is throwing an error and closing the connection, the collaboration plugin will retry infinitely to connect (it's the server that failed).
If the error repeats, we can end up in a infinite loop where the websocket gets connected and disconnected repeatedly.

The current timeout implemetation is based on how many times disconnection happens in a short period of time. If we get disconnected 5 times in less than 2 seconds, the interface will show an error and hide the document.

I personally it is a bit shaky as a solution, but since we don't really have access to the websocket provider implementation, I don't really control the retrial mechanism. For now it is okay, let's see how it plays in the sandbox. If later we need more control on the awareness (for example) it will make sense to fork the plugin.

Another solution would be to play with awareness in some ways, ie. store the error state in here, and avoid closing the connection on error (which is the main reason why the client retries). But as mentioned, it would require to fork the plugin.

Let me know what you think.

### DIFF
--- a/src/components/page/Editor.tsx
+++ b/src/components/page/Editor.tsx
@@ -115,6 +115,7 @@ export function Editor({ item, currentAccount }: Readonly<Props>) {
                           // className="editor-input"
                           // necessary for dnd, for shifting text and show drag icon correctly
                           className={'content-editable-root'}
+                          style={{ fontSize: DEFAULT_FONT_SIZE }}
                           aria-placeholder={'Enter some text...'}
                           placeholder={
                             <div className="editor-placeholder">
--- a/src/components/page/Editor.tsx
+++ b/src/components/page/Editor.tsx
@@ -43,7 +43,7 @@ type Props = {
 
 export function Editor({ item, currentAccount }: Readonly<Props>) {
   const { t } = useTranslation(NS.PageEditor);
-  const { providerFactory, activeUsers, connected, hasTimeout } = useYjs({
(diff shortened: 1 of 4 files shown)


src/components/page/Editor.tsx
```diff
@@ -63,50 +68,75 @@ export function Editor({ item, currentAccount }: Readonly<Props>) {
...
+              }}
+            >
+              {t('RELOAD_BUTTON')}
+            </Button>
+          }
+        >
+          {t('DISCONNECTED_TEXT')}
+        </Alert>
+      )}
+      {/* disable on timeout to prevent further connection attempts */}
```
spaenleh, line 88:
```suggestion
      {/* Unmount component on timeout to prevent further connection attempts */}
```

src/components/page/useYjs.ts
```diff
@@ -68,7 +82,37 @@ export const useYjs = ({ edit }: { edit: boolean }) => {
...
+        setHasTimeout(true);
+      }
+    }
+  }, [disconnectedTimestamps, hasTimeout]);
+
+  return {
+    providerFactory,
+    activeUsers,
+    connected,
+    hasTimeout,
```
spaenleh, line 114: I would call it `hasTimedOut`. But you can choose.

spaenleh (approved):
Looks good.

Should we provide more information about whether the state of the changes have been persisted ? I think they will be lost if we reload the window right ?

PR #1085, opened 2025-08-26
feat: add copy in item page menu

pyphilia:
Add copy item menu in item page. 
Otherwise it is not possible to copy the root (if you come from library for example)

Screenshot 2025-08-26 at 15 24 36


spaenleh (approved):
Looks good !

PR #1086, opened 2025-08-27
fix: remove `waitForMutation` and use `await expect(...).rejects.toThrow()` on throwing mutations

spaenleh:
In this PR I try to address the recurrent issue of failing vitest tests. 

The issue seemed to be linked to a dangling `setTimeout` that is called after test teardown. 
This can happen when using `waitForMutation` in case the test terminates before the timeout. 

I simplified the tests by moving all calls from `mockedMutation.mutate()` to `mockedMutation.mutateAsync()` also removing the unnecessary `act()`.
On mutations that throw because of errors (UNAUTHORIZED, file too large etc...) I added an expect toThrow.

Hopefully this fixes the flackyness of the vitest suite.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -239,7 +239,7 @@ importers:
         version: 19.8.1
       '@cypress/code-coverage':
         specifier: 3.14.5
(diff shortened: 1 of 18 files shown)


pyphilia (approved):
LGTM 🐠

PR #1091, opened 2025-09-02
feat: add link in pages with menu on click

### LINKED ISSUE #1090: Add links feature in pages (text, button, iframe)

pyphilia:
Adds a button to add links.

The menu is available on focus/selected.
Links are alignable (left, right, center, justify)

Links can be
- button (default) 
- embed (iframe or html from iframely)
- text

Note: info for button and embed are fetched on the fly to avoid embedded html. 

https://github.com/user-attachments/assets/707acfa7-b6a9-412a-b0fe-f35f74a63549

Known bugs:
- iframe reload quite often, especially when it switches to "selected"
- cannot click on iframe easily to select it
- button cannot be on the same line as text
- text cannot be on the same line as text. Translate to aligned-links instead of block?
- call to `getLinkMetadata` is not public, so it fails for public users.

close #1090

### DIFF
--- a/src/components/page/BlockWithAlignment.tsx
+++ b/src/components/page/BlockWithAlignment.tsx
@@ -28,7 +28,7 @@ import {
 
 type Props = Readonly<{
   children: ReactNode;
-  format?: ElementFormatType | null;
+  format?: ElementFormatType;
   nodeKey: NodeKey;
   className: Readonly<{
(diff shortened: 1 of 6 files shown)


spaenleh (approved):
Looks good, I have left some comments, I let you decide if you want to tackle them now or later

PR #1104, opened 2025-09-08
fix: allow to add custom tag even when there are hits

### LINKED ISSUE #1101: Cannot add partial tags
Let's suppose I want to add "mathe" as a new tag. It is not possible because it matches with other results.

Image

spaenleh:
I removed the button and instead added an element at the end of the list if there is no exact match:

I also added the number of occurrences as a small chip in the option line.

https://github.com/user-attachments/assets/bda4ac36-a100-4b04-976b-d44bf229cffc

close #1101

### DIFF
--- a/package.json
+++ b/package.json
@@ -192,7 +192,7 @@
     "playwright": "1.52.0",
     "prettier": "3.6.2",
     "storybook": "9.1.5",
-    "typescript": "5.9.2",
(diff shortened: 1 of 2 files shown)


pyphilia (approved):
LGTM, what about having the "Add" option a bit more different from the other options? Like having it in blue, italic, centered?

EDIT: add tests??

PR #1112, opened 2025-09-15
fix: fix link related bugs

pyphilia:
- Toolbar is sticky
- Remove white background for editor, to match with player view 
- Allow click on link (prevent redirection) to select the node and show the options. It is possible to visit the url through the link menu.
- Hide firefox input number's arrow (font size)


spaenleh (approved):
Looks good !

PR #1120, opened 2025-09-17
feat: improve landing menu responsiveness

### LINKED ISSUE #1084: Improve landing menu responsiveness
All the menu items disappear on small screens.
Language switch is not visible

Image

pyphilia:
Show a burger menu when the screen is too small. Actually the screen is too small already at `lg`, because we have 2 menus. I prioritized the action buttons on the right. 

On xs and logged out, the register button is hidden and only the login button is displayed.

https://github.com/user-attachments/assets/29a40cc1-7e9d-4549-a370-db8c4d2009de

close #1084

This PR is based on https://github.com/graasp/client/pull/1119

### DIFF
--- a/src/components/ui/MenuItemLink.tsx
+++ b/src/components/ui/MenuItemLink.tsx
@@ -2,7 +2,12 @@ import { forwardRef } from 'react';
 
 import { MenuItem, MenuItemProps } from '@mui/material';
 
-import { LinkComponent, createLink } from '@tanstack/react-router';
+import {
+  LinkComponent,
+  UseLinkPropsOptions,
+  createLink,
+  useLinkProps,
+} from '@tanstack/react-router';
 
 // eslint-disable-next-line @typescript-eslint/no-empty-object-type
 interface MUIMenuItemProps extends Omit<MenuItemProps, 'href'> {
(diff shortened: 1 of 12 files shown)


src/modules/landing/header/MobileMenu.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  return (
+    <div>
+      <IconButton
+        id="landing-mobile-menu"
+        aria-controls={open ? 'demo-positioned-menu' : undefined}
+        aria-haspopup="true"
+        aria-expanded={open ? 'true' : undefined}
+        onClick={handleClick}
+        sx={{ border: '1px solid' }}
```
spaenleh: Not a big fan of adding the border to the button. I think it looks better without the border.

src/components/ui/MenuItemLink.tsx
```diff
@@ -11,7 +16,15 @@ interface MUIMenuItemProps extends Omit<MenuItemProps, 'href'> {
 
 const MUIMenuItemComponent = forwardRef<HTMLAnchorElement, MUIMenuItemProps>(
   (props, ref) => {
-    return <MenuItem component={'a'} ref={ref} {...props} />;
+    const linkProps = useLinkProps(props as UseLinkPropsOptions);
+
+    // obtain fontweight, usually from activeProps
+    const selected =
+      'selected' in linkProps ? Boolean(linkProps.selected) : false;
```
spaenleh: Not sure where you got that from. When I look at the tanstack/router doc I do not see it in the examples. 
Also in this case we can just spread the props and it works for the selected prop as well.

spaenleh (changes requested):
Changes look good, I would just need some explanations for the changes I highlighted.

PR #1134, opened 2025-10-14
refactor: simplify parents type

pyphilia:
We require only id, name and path from `useParents`, so I simplified the types.

It will allow a simplification in the backend.


spaenleh (approved):
Looks good to me ! Thank you for the work !

